PR #70, opened 2018-06-28
v2

dvoet:


config.json
```diff
@@ -0,0 +1,4 @@
+{
+  "dosResolutionHost": "dcp.bionimbus.org",
+  "bondBaseUrl": ""
```
dvoet: needs to be real

index.js
```diff
@@ -79,8 +80,54 @@ const handler = (req, res) => {
...
+                        res.status(502).send(err);
+                        return;
+                    };
+                    try {
+                        var parsedServiceAccountKey = JSON.parse(bondResponse.text);
+                    } catch(e) {
+                        res.status(500).send(`Service account key not in correct format`);
+                        return;
+                    };
+                    res.status(200).send({'dos': parsedData, 'credentials': parsedServiceAccountKey})
```
dvoet: intent is to return json that looks like
{ "dos": <full response from resolution>, "credentials": <full response from bond>}
I am not certain this does that

docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM node:6.14
+
+## Prepare the image
+ENV DEBIAN_FRONTEND noninteractive
+RUN apt-get update
+RUN apt-get install -y -qq --no-install-recommends unzip
+RUN apt-get clean
```
dvoet: put all the apt-get call into in RUN call: each RUN call results in a separate image so really the cleanup is meaningless on a separate line

same with below and wget, unzip, rm

dvoet:
👍 (github won't let me approve my own PR)

config.json.ctmpl
```diff
@@ -0,0 +1,4 @@
+{
+    "dosResolutionHost": {{if eq (env "ENVIRONMENT") "prod"}}"dcp.bionimbus.org"{{else}}"dcp.bionimbus.org"{{end}},
+    "bondBaseUrl": {{if eq (env "ENVIRONMENT") "prod"}}"broad-bond-prod.appspot.com"{{else if eq (env "ENVIRONMENT") "staging"}}"broad-bond-staging.appspot.com"{{else if eq (env "ENVIRONMENT") "alpha"}}"broad-bond-staging.appspot.com"{{else}}"broad-bond-dev.appspot.com"{{end}}
```
MatthewBemis: if you put `{{with $environment := env "ENVIRONMENT"}}` at the top of the file and `{{end}}` at the bottom you can simplify this line to:

`"bondBaseUrl": "broad-bond-{{$environment}}.appspot.com"`

index.js
```diff
@@ -79,8 +80,53 @@ const handler = (req, res) => {
...
+                        res.status(502).send(bondErr);
+                        return;
+                    }
+                    try {
+                        var parsedServiceAccountKey = JSON.parse(bondResponse.text);
+                    } catch(e) {
+                        res.status(500).send(`Service account key not in correct format`);
+                        return;
+                    }
+                    res.status(200).send({'dos': parsedData, 'credentials': parsedServiceAccountKey["data"]});
```
dvoet: I have changed my mind - we should call credentials googleServiceAccount... I am thinking of the day where we have different types of credentials and it seems easier to name them differently than have a generic "credentials" entry with maybe different types inside.

PR #82, opened 2018-07-23
Cleanup code and use node 8

zarsky-broad:
Also use `package-lock` instead of `npm-shrinkwrap`.


package.json
```diff
@@ -11,16 +11,9 @@
   },
   "dependencies": {
     "ava": "^0.25.0",
-    "axios": "~0.18.0",
     "cors": "^2.8.4",
-    "firebase-functions": "^0.8.2",
-    "nock": "^9.3.3",
-    "npm-shrinkwrap": "^6.1.0",
-    "sinon": "^4.5.0",
+    "sinon": "^6.1.4",
```
gpolumbo-broad: this probably should be moved into `devDependencies`

test/api_adapter.test.js
```diff
@@ -1,12 +1,12 @@
-const test = require("ava");
-const sinon = require("sinon");
-const apiAdapter = require("../api_adapter");
-const superagent = require("superagent");
+const test = require('ava');
+const sinon = require('sinon');
+const { getTextFrom } = require('../api_adapter');
```
gpolumbo-broad: Any reason why you prefer this syntax over something like, `const getTextFrom = require('../api_adapter').getTextFrom` . Is it purely for brevity or is there some other advantage?

zarsky-broad: Good practice. This way multiple members could be destructured out in one place.

PR #85, opened 2018-07-26
Martha v3

zarsky-broad:
I'm not sure how exactly to write tests for this, since it now requires a bearer token to do anything, and the `Promises.all` means method call order is nondeterministic.

### DIFF
--- a/metadata_api.js
+++ b/metadata_api.js
@@ -6,13 +6,10 @@ function getRawMetadata(token, bucket, name) {
         `https://${bucket}.storage.googleapis.com/${encodeURIComponent(name)}`,
         `Bearer ${token}`
     ).catch((e) => {
-        switch (e.status) {
-            case 403:
-                console.error('Permission denied for bucket object');
-                break;
-            default:
-                console.error('Unexpected error while trying to get headers for bucket object')
-        }
+        console.error(e.status === 403 ?
+            'Permission denied for bucket object' :
+            'Unexpected error while trying to get headers for bucket object'
+        );
         throw e;
     });
 }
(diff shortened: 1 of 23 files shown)


service_account_keys.js
```diff
@@ -0,0 +1,34 @@
...
+        auth
+    ).catch((e) => {
+        console.error('Unable to retrieve Service Account Key from Bond');
+        throw e;
+    });
+}
+
+function maybeTalkToSam(auth) {
+    return apiAdapter.getJsonFrom(
+        `${helpers.samBaseUrl()}/api/google/v1/user/petServiceAccount/key`,
```
dvoet: we have had problems with the service account returned here not being able to call the storage json api... we should test early that the signed URL returned using this SA works the way we want

metadata_api.js
```diff
@@ -0,0 +1,80 @@
...
+function getDosObject(dosUri) {
+    const newUri = helpers.dosToHttps(dosUri);
+
+    return apiAdapter.getJsonFrom(newUri).catch((e) => console.error('Failed while trying to retrieve DOS object', e));
+}
+
+async function getMetadata(url, auth, isDos) {
+    try {
+        const uri = isDos ?
+            getGsUriFromDos((await getDosObject(url)).data_object) :
```
dvoet: the auth passed in is not the correct auth to access the gs object returned by dos. Also the dos resolution contains most if not all the information required. Maybe this could be
`isDos ? getDosObjectMetadata : getGsObjectMetadata`

zarsky-broad: The gs object itself is public, so we can use this auth to get its metadata, although not to get a signed url.

service_account_keys.js
zarsky-broad: I can verify that this SA key can't get json metadata (I tried, it will only get xml), but can still get a signed url.

dvoet: getting a signed url is not the same as using a signed url. Anybody can sign a URL, even without access. We know the SA can't get json metadata. Is there the same problem using the singed url.

metadata_api.js
dvoet: no, it is not public... that's why we need to sign it. The one you are currently playing with may be public but the whole point of all this auth business is that they won't be.

zarsky-broad: Ah, understood.

service_account_keys.js
zarsky-broad: Ah, I have tried the signed url's it generates and they have all worked thus far, but do you mean we should validate programmatically? If so, how?

dvoet: that's all, I didn't mean anything fancy

PR #91, opened 2018-08-03
Dev tests passed 1533150774 release [PROD]

MatthewBemis:


dvoet (approved):

gpolumbo-broad (approved):

gpolumbo-broad:
FWIW - we did a lot to try to follow and tune codacy, even though it says we're not up to standards, we're actually way better than we were.  We'll just continue to tweak and adjust code and/or rules.

PR #95, opened 2018-08-28
Forgot to update the dos resolution url in the config.json.ctmpl

gpolumbo-broad:
I had previously only updated `config.json` and forgot to update `config.json.ctmpl`.  Part of me is questioning whether `config.json` should be committed at all or if it should be git ignored.  For now, it's there and it's not hurting anything, but we should consider removing it.


gpolumbo-broad:
I'm pretty sure `config.json` is deleted in this PR, it's just hard to tell because I think it's getting hidden by the `.gitignore`

PR #108, opened 2018-09-21
Fixing config

gpolumbo-broad:


jacmrob (approved):

dvoet (approved):
one thing that struck me a off is that samBaseUrl has https at the front, but the others don't.

PR #112, opened 2018-12-07
circleci changes

marctalbott:
These are the same changes that were made to Bond. Circle will now be responsible for building the image and pushing it to quay

PR #113, opened 2018-12-18
Integration tests for Martha

marctalbott:
[This other PR](https://github.com/broadinstitute/martha/pull/112) hasn't been approved and merged quite yet, but this PR has those changes as well as the integration tests that I wrote.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,7 +56,7 @@ jobs:
       - run:
           name: Run integration tests # this script will change to run the actual tests we want to run
           command: |
-            docker run --entrypoint /bin/bash --workdir "/martha" ${IMAGE_NAME}:${CIRCLE_BRANCH} -c "BASE_URL=\"http://0.0.0.0:8010/dockerized-martha\" npm run-script smoketest"
+            docker run --entrypoint /bin/bash --workdir "/martha" ${IMAGE_NAME}:${CIRCLE_BRANCH} -c "BASE_URL=\"http://0.0.0.0:8010\" npm run-script smoketest"
 
 #      - run:
 #          name: Push to quay # do i need to setup a robot?
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,7 +56,7 @@ jobs:
       - run:
           name: Run integration tests # this script will change to run the actual tests we want to run
           command: |
-            docker run --entrypoint /bin/bash --workdir "/martha" ${IMAGE_NAME}:${CIRCLE_BRANCH} -c "npm run-script smoketest"
+            docker run --entrypoint /bin/bash --workdir "/martha" ${IMAGE_NAME}:${CIRCLE_BRANCH} -c "BASE_URL=\"https://us-central1-broad-dsde-dev.cloudfunctions.net\" npm run-script smoketest"
 
 #      - run:
 #          name: Push to quay # do i need to setup a robot?
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,10 +54,10 @@ jobs:
             docker build -t ${IMAGE_NAME}:${CIRCLE_BRANCH} -f docker/Dockerfile .
             if [[ ${CIRCLE_BRANCH} == "dev" ]]; then docker tag ${IMAGE_NAME}:${CIRCLE_BRANCH} ${IMAGE_NAME}:latest ; fi
 
-      - run:
-          name: Run integration tests # TODO: double check this base_url value
-          command: |
-            docker run --entrypoint /bin/bash --workdir "/martha" ${IMAGE_NAME}:${CIRCLE_BRANCH} -c "BASE_URL=\"https://martha-fiab.dsde-dev.broadinstitute.org:32443/dockerized-martha/us-central1\" npm run-script integration"
+#      - run:
+#          name: Run integration tests
+#          command: |
+#            docker run --entrypoint /bin/bash --workdir "/martha" ${IMAGE_NAME}:${CIRCLE_BRANCH} -c "BASE_URL=\"http://0.0.0.0:8010/dockerized-martha/us-central1\" npm run-script integration"
 
       - run:
           name: Push to quay
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,7 +56,7 @@ jobs:
       - run:
           name: Run integration tests # this script will change to run the actual tests we want to run
           command: |
-            docker run --entrypoint /bin/bash --workdir "/martha" ${IMAGE_NAME}:${CIRCLE_BRANCH} -c "BASE_URL=\"http://0.0.0.0:8010\" npm run-script smoketest"
+            docker run --entrypoint /bin/bash --workdir "/martha" ${IMAGE_NAME}:${CIRCLE_BRANCH} -c "BASE_URL=\"http://0.0.0.0:8010/dockerized-martha/us-central1\" npm run-script smoketest"
 
 #      - run:
 #          name: Push to quay # do i need to setup a robot?
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,7 +56,7 @@ jobs:
       - run:
           name: Run integration tests # this script will change to run the actual tests we want to run
           command: |
-            docker run --entrypoint /bin/bash --workdir "/martha" ${IMAGE_NAME}:${CIRCLE_BRANCH} -c "BASE_URL=\"http://0.0.0.0:8010/dockerized-martha/us-central1\" npm run-script smoketest"
+            docker run --entrypoint /bin/bash --workdir "/martha" ${IMAGE_NAME}:${CIRCLE_BRANCH} -c "BASE_URL=\"http://0.0.0.0:8010/dockerized-martha\" npm run-script smoketest"
 
 #      - run:
 #          name: Push to quay # do i need to setup a robot?
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@
 #
 version: 2
 jobs:
-  npm-test:
+  test:
     docker:
       - image: circleci/node:8.11
 
@@ -34,7 +34,7 @@ jobs:
             git tag $TAG_NAME
             git push origin $TAG_NAME
 
-  integration-test:
+  build-image:
     docker:
       - image: circleci/node:8.11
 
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -34,14 +34,47 @@ jobs:
...
+      - run:
+          name: Build image
+          command: |
+            docker build -t ${IMAGE_NAME}:${CIRCLE_BRANCH} -f docker/Dockerfile .
+            if [[ ${CIRCLE_BRANCH} == "dev" ]]; then docker tag ${IMAGE_NAME}:${CIRCLE_BRANCH} ${IMAGE_NAME}:latest ; fi
+
+      - run:
+          name: Run integration tests # TODO: double check this base_url value
+          command: |
+            docker run --entrypoint /bin/bash --workdir "/martha" ${IMAGE_NAME}:${CIRCLE_BRANCH} -c "BASE_URL=\"https://martha-fiab.dsde-dev.broadinstitute.org:32443/dockerized-martha/us-central1\" npm run-script integration"
```
marctalbott: As the comment above indicates, this is the URL that I was using while testing against my FiaB, but I need to double check that this is going to be the correct URL for circleci to use after building the new image

test/fileSummaryV1/integration_fileSummaryV1.test.js
```diff
@@ -0,0 +1,160 @@
...
+const { GoogleToken } = require('gtoken');
+const { postJsonTo } = require('../../common/api_adapter');
+let unauthorizedToken;
+let authorizedToken;
+
+let fenceAuthLink = 'https://bond-fiab.dsde-dev.broadinstitute.org:31443/api/link/v1/fence/oauthcode?oauthcode=IgnoredByMockProvider&redirect_uri=http%3A%2F%2Flocal.broadinstitute.org%2F%23fence-callback'
+
+test.before(async () => {
+    unauthorizedToken = await new GoogleToken({
+        keyFile: '/Users/mtalbott/ws/broad/martha/firecloud-account.json', // TODO: change filepath to actual .pem key (need it to be .json?)
```
marctalbott: I am not sure where to get this key file from, but it should be accessible from within the FiaB so I just need to determine where it is located and adjust this file path accordingly.

.circleci/config.yml
```diff
@@ -34,14 +34,47 @@ jobs:
...
+      - run:
+          name: Build image
+          command: |
+            docker build -t ${IMAGE_NAME}:${CIRCLE_BRANCH} -f docker/Dockerfile .
+            if [[ ${CIRCLE_BRANCH} == "dev" ]]; then docker tag ${IMAGE_NAME}:${CIRCLE_BRANCH} ${IMAGE_NAME}:latest ; fi
+
+#      - run:
+#          name: Run integration tests
+#          command: |
+#            docker run --entrypoint /bin/bash --workdir "/martha" ${IMAGE_NAME}:${CIRCLE_BRANCH} -c "BASE_URL=\"http://0.0.0.0:8010/dockerized-martha/us-central1\" npm run-script integration"
```
gpolumbo-broad: Should this block be uncommented or removed?

.circleci/config.yml
```diff
@@ -4,7 +4,7 @@
 #
 version: 2
 jobs:
-  build:
+  npm-test:
```
gpolumbo-broad: What do we really want to name this job?  It's building and pushing a docker image to quay (and maybe running tests).  I think we can come up with a name that's more descriptive.  We're running the tests (maybe) just to make sure the image is valid before pushing to quay.  So I think the job is "build-docker-image" or something like that.  And this job may or may not validate the image as part of the build process, but I don't think that needs to be reflected in the name.

package.json
```diff
@@ -9,12 +9,17 @@
...
+    "integration": "ava -m integration*",
+    "integration_v1": "ava -m integration_v1*",
+    "integration_v2": "ava -m integration_v2*",
+    "integration_fileSummaryV1": "ava -m integration_fileSummaryV1*"
   },
   "dependencies": {
     "@google-cloud/storage": "^1.7.0",
     "ava": "^0.25.0",
     "cors": "^2.8.4",
+    "gtoken": "^2.3.0",
```
gpolumbo-broad, line 22: This may cause a domino effect in updating our `npm install` commands, but the `gtoken` module is only needed for dev and testing.  https://medium.com/@dylanavery720/npmmmm-1-dev-dependencies-dependencies-8931c2583b0c . We should see about possibly separating our dependencies accordingly.  That said, maybe this PR isn't the place to do it seeing as we already have several other modules used exclusively for testing that should have been isolated from the start, so the best thing is probably to create an issue to separate them.

test/fileSummaryV1/integration_fileSummaryV1.test.js
```diff
@@ -0,0 +1,162 @@
...
+
+const test = require('ava');
+const supertest = require('supertest')(process.env.BASE_URL);
+const assert = require('assert');
+const { GoogleToken } = require('gtoken');
+const { postJsonTo } = require('../../common/api_adapter');
+let unauthorizedToken;
+let authorizedToken;
+
+let fenceAuthLink = 'https://bond-fiab.dsde-dev.broadinstitute.org:31443/api/link/v1/fence/oauthcode?oauthcode=IgnoredByMockProvider&redirect_uri=http%3A%2F%2Flocal.broadinstitute.org%2F%23fence-callback'
```
gpolumbo-broad: I'm worried about this being a static string.  In particular, being able to change the Bond host that these tests are talking to may be something we need to do.

I've gone back and forth on this before trying to think of a way to ask Martha, "Hey, which Bond instance are you talking to?"  Because this is a configurable thing in Martha.  Which means we probably also want to be able to ask Bond which Sam it is talking to...  

This might be something worth face talking about, so don't worry about changing anything for this PR, but we should talk.

test/fileSummaryV1/integration_fileSummaryV1.test.js
```diff
@@ -0,0 +1,162 @@
...
+
+test.before(async () => {
+    unauthorizedToken = await new GoogleToken({
+        keyFile: 'automation/firecloud-account.pem',
+        email: 'firecloud-dev@broad-dsde-dev.iam.gserviceaccount.com',
+        sub: 'ron.weasley@test.firecloud.org',
+        scope: 'profile email openid https://www.googleapis.com/auth/devstorage.full_control https://www.googleapis.com/auth/cloud-platform'
```
gpolumbo-broad: Try running these tests with just `email` and `openid`.  I think `openid` is the bare minimum we would need.

test/fileSummaryV1/integration_fileSummaryV1.test.js
```diff
@@ -0,0 +1,162 @@
...
+    }).getToken();
+    authorizedToken = await new GoogleToken({
+        keyFile: 'automation/firecloud-account.pem',
+        email: 'firecloud-dev@broad-dsde-dev.iam.gserviceaccount.com',
+        sub: 'hermione.owner@test.firecloud.org',
+        scope: 'profile email openid https://www.googleapis.com/auth/devstorage.full_control https://www.googleapis.com/auth/cloud-platform'
```
gpolumbo-broad: for all the strings passed to `new GoogleToken` extract them to variables.

test/fileSummaryV1/integration_fileSummaryV1.test.js
```diff
@@ -0,0 +1,162 @@
...
+            if (error) { t.log(response.body) };
+            t.end(error);
+        });
+});
+
+test.cb('integration_fileSummaryV1 responds with 200 and file metadata but no signed url if dos uri is valid, but an invalid bearer token is provided', (t) => {
+    supertest
+        .post('/fileSummaryV1')
+        .set('Content-Type', 'application/json')
+        .set('Authorization', `Bearer badToken`)
```
gpolumbo-broad: I think this one and the next are sort of duplicates.  For this test, I suggest you just remove this line and test an otherwise "good" request to `fileSummaryV1` but without a bearer token.

test/fileSummaryV1/integration_fileSummaryV1.test.js
```diff
@@ -0,0 +1,162 @@
...
+            t.end(error);
+        });
+});
+
+test.cb('integration_fileSummaryV1 responds with 502 and unauthorized response if gs uri is valid, but an invalid bearer token is provided', (t) => {
+    supertest
+        .post('/fileSummaryV1')
+        .set('Content-Type', 'application/json')
+        .set('Authorization', `Bearer badToken`)
+        .send({ uri: 'gs://wb-mock-drs-dev/public/dos_test.txt'})
```
gpolumbo-broad: I see you're using different uris in the different tests.  I think when testing different flavors of the same test, it's worth using the same uri, just to be scientific and only change 1 variable at a time.  

That being said, this `gs://` uri reminds me that the Martha logic does different things when it detects a `dos://` uri or a `gs://` uri.  In other words, there are probably several tests in here that we might want to test with each type of uri.  I think these would be limited to the successful `200` responses, but there might be a few of the failure tests that we want to check to.  You might be able to get a sense of what to test if you take a look at the code and look where it diverges based on `gs://` vs. `dos://`.  I'm happy to sit with you too and we can see if we can identify which tests should test both protocols.

test/martha_v1/integration_v1.test.js
```diff
@@ -0,0 +1,40 @@
...
+        .set('Content-Type', 'application/json')
+        .send({ 'url': 'somethingNotValidURL' })
+        .expect((response) => {
+            assert.strictEqual(response.statusCode, 500, "Incorrect status code");
+        })
+        .end((error, response) => {
+            if (error) { t.log(response.body) };
+            t.end(error);
+        });
+});
```
gpolumbo-broad, line 53: add tests for when the pattern is invalid and/or cannot find a match.

gpolumbo-broad (changes requested):

.circleci/config.yml
marctalbott: Yeah, it needs to be removed since the integration tests are going to be run via Jenkins now

gpolumbo-broad (approved):

test/fileSummaryV1/integration_fileSummaryV1.test.js
```diff
@@ -0,0 +1,155 @@
...
+
+let keyFile = 'automation/firecloud-account.pem';
+let serviceAccountEmail = 'firecloud-dev@broad-dsde-dev.iam.gserviceaccount.com';
+let scopes = 'email openid';
+let unauthorizedEmail = 'ron.weasley@test.firecloud.org';
+let authorizedEmail = 'hermione.owner@test.firecloud.org';
+
+let dosUri = 'dos://dg.4503/preview_dos.json';
+let gsUri = 'gs://wb-mock-drs-dev/public/dos_test.txt';
+// TODO: remove static link so bond host can be changed depending on env
```
dvoet: either do this TODO or make a jira for it

dvoet (approved):

test/fileSummaryV1/integration_fileSummaryV1.test.js
marctalbott: Decided to make a ticket for it here: https://broadinstitute.atlassian.net/browse/GAWB-4053

PR #116, opened 2019-01-22
GAWB-4061 No need to use PullApprove at this time

gpolumbo-broad:
We are not using PullApprove anymore because we are not managing/trigging deployments via github merge/commits


dvoet (approved):

gpolumbo-broad:
jenkins retest (let's see if this works for the indie service)

PR #124, opened 2019-06-06
split out smaller helper endpoints

panentheos:
This supports requester pays handling in the UI by limiting the scope of what this service needs to handle. Splits out signed url generation into a separate endpoint that can be called as needed.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -69,5 +69,7 @@ docker run --rm \
      gcloud beta functions deploy martha_v1 --source=. --trigger-http --runtime nodejs8 &&
      gcloud beta functions deploy martha_v2 --source=. --trigger-http --runtime nodejs8 &&
      gcloud beta functions deploy fileSummaryV1 --source=. --runtime nodejs8 --trigger-http &&
+     gcloud beta functions deploy resolveDataObjectUri --source=. --runtime nodejs8 --trigger-http &&
+     gcloud beta functions deploy getSignedUrl --source=. --runtime nodejs8 --trigger-http &&
      npm ci &&
      npm run-script smoketest"
--- a/deploy.sh
+++ b/deploy.sh
@@ -69,7 +69,7 @@ docker run --rm \
      gcloud beta functions deploy martha_v1 --source=. --trigger-http --runtime nodejs8 &&
      gcloud beta functions deploy martha_v2 --source=. --trigger-http --runtime nodejs8 &&
      gcloud beta functions deploy fileSummaryV1 --source=. --runtime nodejs8 --trigger-http &&
-     gcloud beta functions deploy resolveDataObjectUri --source=. --runtime nodejs8 --trigger-http &&
-     gcloud beta functions deploy getSignedUrl --source=. --runtime nodejs8 --trigger-http &&
+     gcloud beta functions deploy resolveDataObjectUriV1 --source=. --runtime nodejs8 --trigger-http &&
+     gcloud beta functions deploy getSignedUrlV1 --source=. --runtime nodejs8 --trigger-http &&
      npm ci &&
      npm run-script smoketest"
--- a/deploy.sh
+++ b/deploy.sh
@@ -69,7 +69,6 @@ docker run --rm \
      gcloud beta functions deploy martha_v1 --source=. --trigger-http --runtime nodejs8 &&
      gcloud beta functions deploy martha_v2 --source=. --trigger-http --runtime nodejs8 &&
      gcloud beta functions deploy fileSummaryV1 --source=. --runtime nodejs8 --trigger-http &&
-     gcloud beta functions deploy resolveDataObjectUriV1 --source=. --runtime nodejs8 --trigger-http &&
      gcloud beta functions deploy getSignedUrlV1 --source=. --runtime nodejs8 --trigger-http &&
      npm ci &&
      npm run-script smoketest"
--- a/handlers/getSignedUrl.js
+++ b/handlers/getSignedUrl.js
@@ -2,7 +2,7 @@ const { Storage } = require('@google-cloud/storage');
 const { getJsonFrom } = require('../common/api_adapter');
 const { bondBaseUrl, promiseHandler, Response, samBaseUrl } = require('../common/helpers');
 
-const getSignedUrl = promiseHandler(async req => {
+const getSignedUrl = promiseHandler(async (req) => {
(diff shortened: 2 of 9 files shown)


handlers/getSignedUrl.js
```diff
@@ -0,0 +1,16 @@
+const { Storage } = require('@google-cloud/storage');
+const { getJsonFrom } = require('../common/api_adapter');
+const { bondBaseUrl, promiseHandler, Response, samBaseUrl } = require('../common/helpers');
+
+const getSignedUrl = promiseHandler(async req => {
+    const { bucket, object, provider } = req.body || {};
```
panentheos: Note: Currently this expects the (optional) provider to be passed directly. Is that what we want, or should we accept the DOS URI and compute the bond provider from that? It depends how formal that mapping is (or is intended to be).

zarsky-broad: Talked to @gpolumbo-broad, tweaked as discussed.

handlers/resolveDataObjectUri.js
```diff
@@ -0,0 +1,8 @@
+const { dataObjectUriToHttps, Response, promiseHandler } = require('../common/helpers');
+
+const resolveDataObjectUri = promiseHandler(async (req) => {
+    const { uri } = req.body || {};
+    return new Response(200, { url: dataObjectUriToHttps(uri) });
+});
+
+module.exports = resolveDataObjectUri;
```
gpolumbo-broad: What's the reason for adding this endpoint?  Is Terra wanting to call and parse the resulting DOS object itself?

handlers/getSignedUrl.js
```diff
@@ -0,0 +1,21 @@
+const { Storage } = require('@google-cloud/storage');
+const { getJsonFrom } = require('../common/api_adapter');
+const { bondBaseUrl, promiseHandler, Response, samBaseUrl, determineBondProvider } = require('../common/helpers');
+
+const getSignedUrl = promiseHandler(async (req) => {
+    const { bucket, object, dataObjectUri } = req.body || {};
+    const auth = req.headers.authorization;
+    const provider = dataObjectUri && determineBondProvider(dataObjectUri);
+    const credentials = provider ?
+        await getJsonFrom(`${bondBaseUrl()}/api/link/v1/${provider}/serviceaccount/key`, auth) :
```
gpolumbo-broad: from line 8, isn't `provider` going to be a boolean?

handlers/getSignedUrl.js
```diff
@@ -0,0 +1,21 @@
...
+        await getJsonFrom(`${samBaseUrl()}/api/google/v1/user/petServiceAccount/key`, auth);
+    const storage = new Storage({ credentials });
+    const [url] = await storage.bucket(bucket).file(object).getSignedUrl({
+        version: 'v4',
+        action: 'read',
+        expires: Date.now() + 36e5
```
gpolumbo-broad: This is basically saying the url never expires, is that what we want?

handlers/getSignedUrl.js
dvoet: where's my compiler?

handlers/resolveDataObjectUri.js
zarsky-broad: Yeah, I think we want to do as much as we can in the UI.

handlers/getSignedUrl.js
zarsky-broad: Here we're using the common js pattern of relying on the fact that `undefined` is false-y.

handlers/getSignedUrl.js
zarsky-broad: For one hour. 1000 * 60 * 60 (ms)

handlers/getSignedUrl.js
zarsky-broad: In js, `"a" && "b"` evaluates to `"b"`

handlers/resolveDataObjectUri.js
gpolumbo-broad: For public data, it shouldn't really be an issue, but for private data that requires Bond to get key information, you'll need to have logic implemented to know how to talk to Bond and which provider to talk to, etc.  I think the goal of Martha was to try to isolate this logic.  I will invoke the @dvoet to make sure he's aware.

panentheos: Centralizing that logic in Martha continues to be a goal. See `getSignedUrl` above.

deploy.sh
```diff
@@ -69,5 +69,7 @@ docker run --rm \
      gcloud beta functions deploy martha_v1 --source=. --trigger-http --runtime nodejs8 &&
      gcloud beta functions deploy martha_v2 --source=. --trigger-http --runtime nodejs8 &&
      gcloud beta functions deploy fileSummaryV1 --source=. --runtime nodejs8 --trigger-http &&
+     gcloud beta functions deploy resolveDataObjectUri --source=. --runtime nodejs8 --trigger-http &&
+     gcloud beta functions deploy getSignedUrl --source=. --runtime nodejs8 --trigger-http &&
```
dvoet: these need version indicators

handlers/resolveDataObjectUri.js
dvoet: spreading the logic to parse the json response from a dos/drs resolution is something martha aims to avoid. The spec is unstable and we want to be able to handle multiple versions of the spec in 1 place, which is martha. Exposing this url goes against martha's purpose.

gpolumbo-broad:
@panentheos @zarsky-broad - I assume that requester pays needs to work when localizing files in Cromwell too.  If we start breaking out logic in the UI, I'm assuming we'll need to reimplement that logic here in Martha too or in Cromwell?   Is there a way we could modifying `fileSummaryV1` in a non-breaking way or implement a new `fileSummaryV2` or `martha_v3`?  I am not very familiar with how Requester Pays works, so my apologies if there's something I may be missing.

handlers/resolveDataObjectUri.js
panentheos: What's the abstraction we want to provide, then? DOS => metadata+gsuri?

panentheos:
My understanding was that Cromwell was already handling requester pays buckets. Is that not true?

panentheos:
This is is ready for re-review.

gpolumbo-broad (approved):

dvoet (approved):

PR #125, opened 2019-06-14
For HCA return only the cloud native URL

### LINKED ISSUE #982: Broad: Change Martha to bypass request to Bond for HCA DRS URIs
Related to: https://github.com/DataBiosphere/azul/issues/981

For more info, see: [Terra/Cromwell Access to Files Identified by DRS URI for the HCA/Terra MVP](https://docs.google.com/document/d/18WDYr3uS82Eu1GhHmeH6PX7tFY6Zk8piKd3ixOoE8Bg/edit#heading=h.9strq6o3dhsc)

Owned by Greg Polumbo (Broad)

Done when:
The Broad's Martha component has been updated to request the HCA provider for HCA DRS URIs

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-611)
┆Project Name: azul
┆Issue Number: AZUL-611

### LINKED ISSUE #1090: Broad: Terra/Cromwell/Martha does not work for DRS URIs that resolve to cloud native URLs that are accessible without any additional credentials
Terra/Cromwell as it stands does not appear to work for DRS URIs that resolve to cloud-native URLs and are accessible without any additional credentials, either because the cloud-native URLs are publicly accessible or are accessible by the Terra user’s Proxy Group.

Excerpt from email describing this issue:

Reviewing the workflow log, I noticed that the dev deployment of the HCA Data Browser (https://dev.data.humancellatlas.org/explore/projects) is using the HCA DSS staging deployment (intentionally, and for good reason).
So, I added the Terra/FC Proxy Groups to the HCA DSS Google checkout bucket for staging instead (“org-humancellatlas-dss-checkout-staging”).
I then tried rerunning the same md5sum workflow configuration from the same Terra workspace … and got the same result, failure, with the same log content as before.

Looking for other possible problems, I again verified that Martha is returning the correct Google cloud native URL for the DRS URI in the Terra workspace, which it is:

 $ curl --request POST \
>   --url https://us-central1-broad-dsde-prod.cloudfunctions.net/martha_v2 \
>   --header 'content-type: application/json' \
>   --data '{ "url": "dos://service.dev.explore.data.humancellatlas.org/7c800467-9143-402f-b965-4e7cad75c1e6?version=2019-05-26T130511.722646Z" }'
{"dos":{"data_object":{"id":"7c800467-9143-402f-b965-4e7cad75c1e6","urls":[{"url":"https://service.dev.explore.data.humancellatlas.org/dss/files/7c800467-9143-402f-b965-4e7cad75c1e6?version=2019-05-26T130511.722646Z&replica=aws&wait=1&fileName=merged.bam"},{"url":"gs://org-humancellatlas-dss-checkout-staging/blobs/8713ae74c6378311f1f3f0c2ead08a2402d90035d7823bbbc2d68d147f898789.06af02fe3b4bf718cd7d4ec35e3f624387548b86.a34f51039b67d4bf3418c5ad1b94a1ea.c6c25e15"}],"size":"22403106","checksums":[{"checksum":"8713ae74c6378311f1f3f0c2ead08a2402d90035d7823bbbc2d68d147f898789","type":"sha256"}],"aliases":["merged.bam"],"version":"2019-05-26T130511.722646Z","name":"merged.bam"}}}

I then checked what happens when Martha is invoked with a bearer token, as it is from Cromwell.
Although I just used a bogus bearer token value, the results seem informative:

$   curl --request POST \
>   --url https://us-central1-broad-dsde-prod.cloudfunctions.net/martha_v2 \
>   --header 'content-type: application/json' \
>   --header 'Authorization: Bearer ABC123' \
>   --data '{ "url": "dos://service.dev.explore.data.humancellatlas.org/7c800467-9143-402f-b965-4e7cad75c1e6?version=2019-05-26T130511.722646Z" }'
{"status":401,"response":{"req":{"method":"GET","url":"broad-bond-prod.appspot.com/api/link/v1/dcf-fence/serviceaccount/key","headers":{"user-agent":"node-superagent/3.8.3","authorization":"Bearer ABC123"}},"header":{"content-type":"application/json","x-cloud-trace-context":"23a2ab1fc041e2219401d9b3adcac1b7;o=1","date":"Thu, 13 Jun 2019 13:22:17 GMT","server":"Google Frontend","content-length":"204","connection":"close"},"status":401,"text":"{\n \"error\": {\n  \"code\": 401, \n  \"errors\": [\n   {\n    \"domain\": \"global\", \n    \"message\": \"Invalid authorization token\", \n    \"reason\": \"required\"\n   }\n  ], \n  \"message\": \"Invalid authorization token\"\n }\n}"}}

It looks like if a bearer token is provided in the request to Martha, Martha tries getting credentials from the “dcf-fence” Bond provider, and in this case, fails.
This is consistent with the behavior I expected from reading the Martha code (and why I gave this less than a 50/50 chance yesterday :).

So, it seems some tweaks are needed to the Broad Martha/Bond plumbing to handle the case of DRS returning a cloud native URL that is accessible without additional credentials / service account key. This case may legitimately arise either because the cloud native URL is publicly accessible, or is in a bucket that the Terra user’s service account has access to.

If my interpretation is correct, this needs to be fixed.

The name of the workspace I have been using for these experiments is "mbaumann test workflow no dash headers 20190612_2029” and I have added Alex, Brad and Brian O. as Owners.
Feel free to share it with whomever else you like to further the investigation. Again, it is all public data.

I have also added "GROUP_All_Users@firecloud.org” to the “org-humancellatlas-dss-checkout-staging” bucket with the “Storage Object Viewer” role, so theoretically you all should be able to access the data in it.

At this point, I think I have done all I can and the ball is in the Broad’s court.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-681)
┆Project Name: azul
┆Issue Number: AZUL-681

mikebaumann:
Currently, all Human Cell Atlas (HCA) is open access, and
no user-specific access control is performed.
The HCA checkout buckets, from which the data is accessed,
grant object read permission to GROUP_All_Users@firecloud.org,
so no additional credentials are required or available.
Therefore, for HCA data access, return only
the cloud-native URL, and bypass the request to Bond.
Both the `martha_v2` and `fileSummaryV1` endpoints have been updated to support this behavior.

The new `getSignedUrlV1` endpoint has been updated to obtain a user pet SA key from SAM for HCA DRS URIs.

The unit tests have been augmented to cover the "happy path" through all functional code changes, and most other paths are already covered by existing tests.

For more information, please see: [Terra/Cromwell Access to Files Identified by DRS URIs for the HCA/Terra MVP](https://docs.google.com/document/d/18WDYr3uS82Eu1GhHmeH6PX7tFY6Zk8piKd3ixOoE8Bg)

Resolves DataBiosphere/azul#982.
Partially resolves DataBiosphere/azul#1090, as changes to Cromwell are also needed for that one.

### DIFF
--- a/common/helpers.js
+++ b/common/helpers.js
@@ -76,6 +76,7 @@ function dataObjectUriToHttps(dataObjectUri) {
 const BondProviders = Object.freeze({
     FENCE: 'fence',
     DCF_FENCE: 'dcf-fence',
+    HCA: 'hca', // Human Cell Atlas
     get default() {
         return this.DCF_FENCE;
     }
@@ -85,6 +86,8 @@ function determineBondProvider(urlString) {
     const url = URL.parse(urlString);
     if (url.host === 'dg.4503') {
         return BondProviders.FENCE;
+    } else if (url.host.endswith('.humancellatlas.org')) {
+        return BondProviders.HCA;
     } else {
         return BondProviders.default;
     }
--- a/common/helpers.js
+++ b/common/helpers.js
@@ -86,7 +86,7 @@ function determineBondProvider(urlString) {
     const url = URL.parse(urlString);
     if (url.host === 'dg.4503') {
         return BondProviders.FENCE;
-    } else if (url.host.endswith('.humancellatlas.org')) {
+    } else if (url.host.endsWith('.humancellatlas.org')) {
         return BondProviders.HCA;
     } else {
         return BondProviders.default;
--- a/handlers/getSignedUrlV1.js
+++ b/handlers/getSignedUrlV1.js
@@ -1,5 +1,5 @@
 const { Storage } = require('@google-cloud/storage');
-const { getJsonFrom } = require('../common/api_adapter');
+const apiAdapter = require('../common/api_adapter');
 const { bondBaseUrl, promiseHandler, Response, samBaseUrl, determineBondProvider, BondProviders } = require('../common/helpers');
 
 const getSignedUrlV1 = promiseHandler(async (req) => {
@@ -8,10 +8,10 @@ const getSignedUrlV1 = promiseHandler(async (req) => {
     const provider = dataObjectUri && determineBondProvider(dataObjectUri);
     try {
         const credentials = (provider && provider !== BondProviders.HCA) ?
-            (await getJsonFrom(`${bondBaseUrl()}/api/link/v1/${provider}/serviceaccount/key`, auth)).data :
-            await getJsonFrom(`${samBaseUrl()}/api/google/v1/user/petServiceAccount/key`, auth);
+            (await apiAdapter.getJsonFrom(`${bondBaseUrl()}/api/link/v1/${provider}/serviceaccount/key`, auth)).data :
+            await apiAdapter.getJsonFrom(`${samBaseUrl()}/api/google/v1/user/petServiceAccount/key`, auth);
         const storage = new Storage({ credentials });
-        const [url] = await storage.bucket(bucket).file(object).getSignedUrl({
+        const url = await storage.bucket(bucket).file(object).getSignedUrl({
             version: 'v4',
             action: 'read',
             expires: Date.now() + 36e5
--- a/handlers/getSignedUrlV1.js
+++ b/handlers/getSignedUrlV1.js
@@ -1,13 +1,13 @@
 const { Storage } = require('@google-cloud/storage');
 const { getJsonFrom } = require('../common/api_adapter');
-const { bondBaseUrl, promiseHandler, Response, samBaseUrl, determineBondProvider } = require('../common/helpers');
+const { bondBaseUrl, promiseHandler, Response, samBaseUrl, determineBondProvider, BondProviders } = require('../common/helpers');
 
 const getSignedUrlV1 = promiseHandler(async (req) => {
     const { bucket, object, dataObjectUri } = req.body || {};
     const auth = req.headers.authorization;
     const provider = dataObjectUri && determineBondProvider(dataObjectUri);
     try {
-        const credentials = provider ?
+        const credentials = (provider && provider !== BondProviders.HCA) ?
             (await getJsonFrom(`${bondBaseUrl()}/api/link/v1/${provider}/serviceaccount/key`, auth)).data :
             await getJsonFrom(`${samBaseUrl()}/api/google/v1/user/petServiceAccount/key`, auth);
         const storage = new Storage({ credentials });
--- a/handlers/getSignedUrlV1.js
+++ b/handlers/getSignedUrlV1.js
@@ -11,7 +11,7 @@ const getSignedUrlV1 = promiseHandler(async (req) => {
             (await apiAdapter.getJsonFrom(`${bondBaseUrl()}/api/link/v1/${provider}/serviceaccount/key`, auth)).data :
             await apiAdapter.getJsonFrom(`${samBaseUrl()}/api/google/v1/user/petServiceAccount/key`, auth);
         const storage = new Storage({ credentials });
-        const url = await storage.bucket(bucket).file(object).getSignedUrl({
+        const [url] = await storage.bucket(bucket).file(object).getSignedUrl({
             version: 'v4',
             action: 'read',
             expires: Date.now() + 36e5
--- a/martha_v2/martha_v2.js
+++ b/martha_v2/martha_v2.js
@@ -13,7 +13,7 @@ function maybeTalkToBond(req, provider = BondProviders.default) {
     let myPromise;
     // Currently HCA data access does not require additional credentials.
     // The HCA checkout buckets allow object read access for GROUP_All_Users@firecloud.org.
-    if (req && req.headers && req.headers.authorization && provider != BondProviders.HCA) {
+    if (req && req.headers && req.headers.authorization && provider !== BondProviders.HCA) {
         myPromise = apiAdapter.getJsonFrom(`${bondBaseUrl()}/api/link/v1/${provider}/serviceaccount/key`, req.headers.authorization);
     } else {
         myPromise = Promise.resolve();
--- a/martha_v2/martha_v2.js
+++ b/martha_v2/martha_v2.js
@@ -11,7 +11,9 @@ function parseRequest(req) {
 
 function maybeTalkToBond(req, provider = BondProviders.default) {
     let myPromise;
-    if (req && req.headers && req.headers.authorization) {
+    // Currently HCA data access does not require additional credentials.
+    // The HCA checkout buckets allow object read access for GROUP_All_Users@firecloud.org.
+    if (req && req.headers && req.headers.authorization && provider != BondProviders.HCA) {
         myPromise = apiAdapter.getJsonFrom(`${bondBaseUrl()}/api/link/v1/${provider}/serviceaccount/key`, req.headers.authorization);
     } else {
         myPromise = Promise.resolve();
--- a/fileSummaryV1/service_account_keys.js
+++ b/fileSummaryV1/service_account_keys.js
@@ -3,15 +3,19 @@ const apiAdapter = require('../common/api_adapter');
 
 function maybeTalkToBond(auth, url) {
     const provider = determineBondProvider(url);
-    return apiAdapter.getJsonFrom(
-        `${bondBaseUrl()}/api/link/v1/${provider}/serviceaccount/key`,
-        auth
-    ).then(
-        (res) => res.data
-    ).catch(() => {
-        console.error(new Error('Unable to retrieve Service Account Key from Bond'));
+    if (provider === BondProviders.HCA) {
         return Promise.resolve();
-    });
+    } else {
+        return apiAdapter.getJsonFrom(
+            `${bondBaseUrl()}/api/link/v1/${provider}/serviceaccount/key`,
+            auth
+        ).then(
+            (res) => res.data
+        ).catch(() => {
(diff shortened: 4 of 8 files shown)


common/helpers.js
```diff
@@ -76,6 +76,7 @@ function dataObjectUriToHttps(dataObjectUri) {
 const BondProviders = Object.freeze({
     FENCE: 'fence',
     DCF_FENCE: 'dcf-fence',
+    HCA: 'hca', // Human Cell Atlas
```
panentheos, line 79: This implies that there is a bond provider for HCA, which according to the description there isn't. It might be clearer to remove this, return `null` from `determineBondProvider`, and make the bond call conditional on the existence of the provider string.

martha_v2/martha_v2.js
```diff
@@ -11,7 +11,9 @@ function parseRequest(req) {
 
 function maybeTalkToBond(req, provider = BondProviders.default) {
     let myPromise;
-    if (req && req.headers && req.headers.authorization) {
+    // Currently HCA data access does not require additional credentials.
+    // The HCA checkout buckets allow object read access for GROUP_All_Users@firecloud.org.
+    if (req && req.headers && req.headers.authorization && provider != BondProviders.HCA) {
```
panentheos: In addition to here, we'll also need a similar change in `fileSummaryV1/service_account_keys.js`

common/helpers.js
gpolumbo-broad: At first glance, I agreed with you @panentheos , but now that I had a chance to talk this over with @abaumann and @geoffjentry , I'm OK with it.  Eventually, HCA will have a real provider.  Due to limitations in the DOS spec at the moment, it's hard to know which provider to use for which DOS URIs.  This at least keeps all that logic isolated to Martha.

gpolumbo-broad (approved):
@dvoet - FYI this PR.  Had a meeting with Jeff and Alex and we talked this through.  I can fill you in if you have additional questions.

common/helpers.js
mikebaumann: @panentheos, @gplumbo-broad Thank you both for your review and comments.
Indeed, this is just a first, tiny step towards a bona fide Bond provider for HCA that will be developed over time. The significance of the Bond provider to the Terra/HCA interaction is clearly evident in this design document [Terra/Cromwell Access to Files Identified by DRS URIs for the HCA/Terra MVP](https://docs.google.com/document/d/18WDYr3uS82Eu1GhHmeH6PX7tFY6Zk8piKd3ixOoE8Bg)

The immediate objective is to enable Terra to access HCA for the HCA handoff to Terra MVP initiative. Currently, all HCA data is open/public access, and anyone can obtain a presigned URL for any data. However, Cromwell currently requires cloud-native URLs, so HCA is providing special access to the Broad (Mint and Terra) via cloud-native URLs.  HCA will be supporting controlled access data with user-specific authN/authZ over time and is developing the Fusillade system to support this.
Although currently, HCA data is open/public access, the bucket containing it is access controlled. For the immediate/near-term, the recent plan had been to develop a minimal HCA provider based on the `mock-provider` to provide a single/fixed HCA service account key for all Terra users. Then @abaumann proposed the simpler, easier short-term solution of simply adding  `GROUP_All_Users@firecloud.org` with read access to the ACL of the HCA bucket, ... and hence this PR. :)

dvoet (approved):

martha_v2/martha_v2.js
gpolumbo-broad: Good catch @panentheos, thanks.  @mikebaumann - I think we're looking at around this line: https://github.com/mikebaumann/martha/blob/764ec6a1bb3f39349165b113facb0c2932069297/fileSummaryV1/service_account_keys.js#L5 and doing something similar which is if the provider is `HCA`, then just return a resolved Promise.

martha_v2/martha_v2.js
```diff
@@ -11,7 +11,9 @@ function parseRequest(req) {
 
 function maybeTalkToBond(req, provider = BondProviders.default) {
     let myPromise;
-    if (req && req.headers && req.headers.authorization) {
+    // Currently HCA data access does not require additional credentials.
+    // The HCA checkout buckets allow object read access for GROUP_All_Users@firecloud.org.
+    if (req && req.headers && req.headers.authorization && provider !== BondProviders.HCA) {
```
gpolumbo-broad, line 16: https://github.com/mikebaumann/martha/blob/764ec6a1bb3f39349165b113facb0c2932069297/fileSummaryV1/service_account_keys.js#L5 

We need to do something similar which is if the provider is HCA, then just return a resolved Promise.

This will ensure that behavior in both endpoints, `martha_v2` and `fileSummaryV1` remain consistent.

gpolumbo-broad (changes requested):

martha_v2/martha_v2.js
mikebaumann: Thanks, @panentheos, @gpolumbo-broad!
I'm back in the office now and committed the corresponding change for File Summary V1. Please have a look. If you would like the conditional to be structured differently, please let me know.

martha_v2/martha_v2.js
mikebaumann: This change has now been made, as per my recent comment above.

mikebaumann:
@gpolumbo-broad Unit tests have been added for all HCA-related functional changes to Martha. 
I think this PR is now done and ready to merge, pending your review.
Thanks!

panentheos:
There's one more change needed here. There was a new endpoint added on another branch (`getSignedUrlV1`) that will also need to know about the HCA special case. (A rebase should pick up the new code)

@gpolumbo-broad Note that this is necessary because `determineBondProvider` doesn't return null.

mikebaumann:
Thanks, @panentheos. Yes, I think the change you made will be easy to update for the HCA case, and once that is done, it should work as intended/expected for HCA.

Will you be adding tests for the change you made? If so, I will update those as well.

Rebasing my branch pulled in your changes just fine. I expect to update the HCA PR for your latest change later today.

mikebaumann:
I pushed a commit to make the necessary HCA-related change for `getSignedUrlV1`.
I then started looking at adding unit tests for this change, and 
a question arose.
@panentheos, regarding your change for getSignedUrlV1, it seems moving the existing `fileSummaryV1.createSignedGsUrl` code into `common.helpers.js` and using that to
implement `handlers.getSignedUrlV1` would reduce code duplication, help ensure
code consistency, and make mocking for testing getSignedUrlV1 easier.
Is there a reason you didn't do that?
Although I suppose there is a way to mock the following in ava/sinon:
```
storage.bucket(bucket).file(object).getSignedUrl
```
mocking the following single call seems a lot easier: :-)
```
createSignedGsUrl
```

panentheos:
@mikebaumann The bulk of `createSignedGsUrl` is just plumbing, and the only abstraction of any interest is the expiration timestamp. And while happen to be using the same timestamp in `getSignedUrlV1`, there's no requirement that they be the same as each other going forward. So to me it doesn't seem substantial enough to justify the cost of tying together two otherwise independent functions.

mikebaumann:
@panentheos Yes, the bulk of `createSignedGsUrl` is just plumbing, yet it happens to be the
same plumbing as is used in `getSignedUrlV1`. That's fine, it is a borderline case for refactoring,
and my main motivation for asking was to simplify unit testing of `getSignedUrlV1`.
Happily, I figured out how to test the current code using ava/sinon, although minor changes were required to `getSignedUrlV1.js` to enable testing with sinon (AFAIK), and these changes are consistent with similar code in the `fileSummaryV1` and `martha_v2` endpoints.

Thanks in advance for your review.

mikebaumann:
@gpolumbo-broad, @panentheos Please review this PR again.
The new `getSignedUrlV1` endpoint has been updated for HCA, and basic unit tests were added for the same.
I believe all comments have been addressed/resolved, and hopefully, this will be the final review.

handlers/getSignedUrlV1.js
```diff
@@ -1,17 +1,17 @@
...
     try {
-        const credentials = provider ?
-            (await getJsonFrom(`${bondBaseUrl()}/api/link/v1/${provider}/serviceaccount/key`, auth)).data :
-            await getJsonFrom(`${samBaseUrl()}/api/google/v1/user/petServiceAccount/key`, auth);
+        const credentials = (provider && provider !== BondProviders.HCA) ?
+            (await apiAdapter.getJsonFrom(`${bondBaseUrl()}/api/link/v1/${provider}/serviceaccount/key`, auth)).data :
+            await apiAdapter.getJsonFrom(`${samBaseUrl()}/api/google/v1/user/petServiceAccount/key`, auth);
         const storage = new Storage({ credentials });
-        const [url] = await storage.bucket(bucket).file(object).getSignedUrl({
+        const url = await storage.bucket(bucket).file(object).getSignedUrl({
```
panentheos: This doesn't look right. The library call returns a single element array that needs to be destructured.

mikebaumann: Thanks! True, indeed:
```
export declare type GetSignedUrlResponse = [string];
export interface GetSignedUrlCallback {
    (err: Error | null, url?: string): void;
}
```

Albeit unexpected, at least to me.
I restored the original line in 'getSignedUrlV1` and updated the unit test mock value to be an array of string instead of string.

panentheos (commented):
Looks good to me, I'll let someone on CA give the official approval.

mikebaumann:
Thanks very much for the review, @panentheos!

gpolumbo-broad (approved):

PR #130, opened 2019-08-29
Gpolumbo case sensitivity

gpolumbo-broad:


gpolumbo-broad:
Merging into other branch which will get the important PR

PR #132, opened 2019-09-03
CA-388 Adding some commented/pending DOS URIs to Live Tests

gpolumbo-broad:
When we originally added support for humancellatlas.org DOS URIs, we got a list of some DOS URIs we could use for testing.  These were sitting around in an uncommitted branch on my machine and I did not want to lose them, so I have checked them in here.

They are commented out because at this time, humancellatlas.org does not have an identity provider for protecting the data, it should be publicly available but they also don't want their data sitting in a publicly open bucket.  Therefore, when calling `martha_v2` or `fileSummaryV1` in Martha to resolve a `*.humancellatlas.org` URL, we bypass `talkToBond` step and just resolve the DOS URI regardless of whether you provided a `Bearer` token or not.  This is different behavior from the way Martha normally does things and therefore the `live_test`s using `*.humancellatlas.org` DOS URIs are currently failing.  

I am creating this PR so we do not lose this work.  The right thing to do would be to fix the `live_test`s so that they work properly with `*.humancellatlas.org` URIs, but I do not have the time to fix them right this second.


gpolumbo-broad:
Old PR - closing this out

PR #136, opened 2020-02-04
improve error handling by logging full details [WA-91]

rsasch:
- added catches to api calls
- reverted new Error() calls

closes https://broadworkbench.atlassian.net/browse/WA-91


rsasch:
re-naming branch and re-creating PR because it might be causing issues

PR #137, opened 2020-02-05
add minnie-kenny [WA-96]

rsasch:
Closes https://broadworkbench.atlassian.net/browse/WA-96


rsasch:
re-naming branch and re-creating PR because it might be causing issues

PR #140, opened 2020-02-10
Retry 5xx and 429 errors with backoff [WA-90]

salonishah11:
JIRA issue [here](https://broadworkbench.atlassian.net/browse/WA-90).


kshakir (approved):
LGTM. I'm happy when the linter is happy

rsasch (approved):
Ditto — looks like a small change to appease the linter and then you should be all set 👍

PR #144, opened 2020-02-21
changed dos uri in request for failing test [WA-125]

rsasch:


rsasch:
Issue was found to be with changes made to Bond response codes.

PR #145, opened 2020-03-03
WA-125 Modify Martha integration tests to not rely on particular status codes from Bond.

wchamber:
Martha does not care what error code Bond returns other than these integration tests. We can have the tests be less brittle to changes in error codes and just verify that the requests fail in these casees. Fixing this now as these integration tests are failing due to changes in Bond error codes.


gpolumbo-broad (approved):

aednichols (approved):
Thank you!

PR #150, opened 2020-04-01
Better error handling and additional logging [WA-128]

salonishah11:
This PR closes: https://broadworkbench.atlassian.net/browse/WA-128

In case that Martha receives an empty JSON body from Bond or the external DRS Resolver, it will catch it, and create a new response with status 500 so that it can be retired. The final response that is returned upstream will be this (for now), which is similar to the format when Martha receives 401 or 404. 
```
{
    "status": 500,
    "data": {
        "response": {
            "status": 500,
            "text": "Something went wrong while trying to resolve url '127.0.0.1:8080/api/link/v1/dcf-fence/serviceaccount/key'. It came back with empty JSON body!"
        }
    }
}
```
This response should be standardized in `martha_v3` as part of ticket [Martha V3: Decide and implement a standard response schema](https://broadworkbench.atlassian.net/browse/WA-160).

### DIFF
--- a/common/api_adapter.js
+++ b/common/api_adapter.js
@@ -1,4 +1,5 @@
 const request = require('superagent');
+const {Response} = require('../common/helpers');
 
 const MAX_RETRY_ATTEMPTS = 5;
 const INITIAL_BACKOFF_DELAY = 1000;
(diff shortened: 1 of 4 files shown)


common/api_adapter.js
```diff
@@ -32,10 +33,32 @@ async function getHeaders(url, authorization) {
...
         const {body} = await get('get', url, authorization);
-        return body;
+
+        /*
+         handle the case when Martha receives empty JSON body. The reason behind forming a response with status 500
+         and throwing it in `try` is so that it can be caught locally and be retried.
+         */
+        if (Object.keys(body).length === 0) {
+            console.log(`Received an empty JSON body while trying to resolve url '${url}'. Creating a response with ` +
+            'status 500 and will retry.');
```
kshakir: Is a retry guaranteed? I assume there are not infinite retries, and will at some point stop.

Alternatively, we could just log the `retryAttempt`, say:

```
`status 500 during attempt ${retryAttempt}`
```

salonishah11: >Is a retry guaranteed? I assume there are not infinite retries, and will at some point stop.

yes. It will be caught locally, and that [catch block](https://github.com/broadinstitute/martha/blob/942005e4a869c34cdb51f3ba626c929290ddbf6e/common/api_adapter.js#L41-L55) has retry logic which also checks the retry attempt. So it should not be retried infinite times. The idea is that if receiving an empty json body was just once and not consistent, retrying would help in this case. We can certainly add the retry attempt in the log message.
>Alternatively, we could just log the retryAttempt

We can also add that. I will add it in the log message

salonishah11: I misunderstood the comment 🤦‍♀. Yes it does make more sense to log the retry attempt instead of saying `will retry` every time as that would not be true on the last attempt.

rsasch (approved):
👍

kshakir (approved):

PR #153, opened 2020-04-28
Jade Data Repo integration DRS v1.0 prefix [WA-178]

salonishah11:
This PR introduces new version of martha, `martha_v3`. Currently in the codebase, there are 2 functions common across `martha_v2` and `martha_v3` handlers namely `maybeTalkToBond` and `aggregateResponses` which I haven't refactored. Since they both will change in the eventual tickets for JDR support (auth header and standard response), I have left them as it is. Same applies to the tests in `martha_v3.test.js` where mock responses and assertion checks will eventually change.

Closes https://broadworkbench.atlassian.net/browse/WA-178.

### DIFF
--- a/common/helpers.js
+++ b/common/helpers.js
@@ -2,10 +2,23 @@ const url = require('url');
 const config = require('../config.json');
 
 const dataGuidsHostPrefix = 'dg.';
-const dataObjectPathPrefix = '/ga4gh/dos/v1/dataobjects/';
+const dosDataObjectPathPrefix = '/ga4gh/dos/v1/dataobjects/';
+const drsDataObjectPathPrefix = '/ga4gh/drs/v1/objects/';
+const jadeDataRepoHostRegex = /jade(-terra)?.datarepo-(dev|prod).broadinstitute.org/;
+
+// Regex drops any leading or trailing "/" characters and gives the path out of capture group 1
+const pathSlashRegex = /^\/?([^/]+.*?)\/?$/;
+
+const samBaseUrl = () => config.samBaseUrl;
+const jadeDataRepoUrl = () => config.jadeDataRepoBaseUrl;
+
 
 function hasDataGuidsHost(someUrl) {
     return someUrl.host.startsWith(dataGuidsHostPrefix);
+}
+
+function hasJadeDataRepoHost(someUrl) {
+    return someUrl.host === jadeDataRepoUrl();
 }
 
 function isDataGuidsUrl(someUrl) {
@@ -13,13 +26,23 @@ function isDataGuidsUrl(someUrl) {
 }
 
 function validateDataObjectUrl(someUrl) {
-    if (hasDataGuidsHost(someUrl) && !someUrl.pathname) {
-        throw new Error(`Data Object URIs with '${dataGuidsHostPrefix}*' host are required to have a path: "${url.format(someUrl)}"`);
+    const hasJDRHost = hasJadeDataRepoHost(someUrl);
+
+    if ((hasDataGuidsHost(someUrl) || hasJDRHost) && !someUrl.pathname) {
+        throw new Error(`Data Object URIs with either '${dataGuidsHostPrefix}*' or '${jadeDataRepoUrl()}' as host are required to have a path: "${url.format(someUrl)}"`);
+    }
+
+    /*
+        This is to notify the user that they have passed data object with a JDR host not supported in this env. If we don't check this condition,
+        then in `determinePathname` function, since the host will not match to the JDR host supported by this env, it will form the HTTPS url with
+        dosDataObjectPathPrefix i.e https://jade-data-repo-host/ga4gh/dos/v1/dataobjects/url-here, and contact JDR eventually which will result in
+        an error. And it might be difficult for the user to understand the cause of this error.
+     */
+    if (jadeDataRepoHostRegex.test(someUrl.host) && !hasJDRHost) {
+        throw new Error(`Data Object URI belongs to a different Jade Data Repo host. This version supports '${jadeDataRepoUrl()}'. URL passed: '${url.format(someUrl)}'`)
     }
 }
 
-// Regex drops any leading or trailing "/" characters and gives the path out of capture group 1
-const pathSlashRegex = /^\/?([^/]+.*?)\/?$/;
 /**
  *  Filter off the null entries first (because `regex.exec(null)` is TRUTHY, because of course it is)
  *  Then run the regex to get the path part without leading or trailing slashes
@@ -41,9 +64,11 @@ function determineHostname(someUrl) {
 
 function determinePathname(someUrl) {
     if (isDataGuidsUrl(someUrl)) {
-        return constructPath([dataObjectPathPrefix, someUrl.hostname, someUrl.pathname]);
+        return constructPath([dosDataObjectPathPrefix, someUrl.hostname, someUrl.pathname]);
+    } else if (hasJadeDataRepoHost(someUrl)) {
+        return constructPath([drsDataObjectPathPrefix, someUrl.pathname])
     } else {
-        return constructPath([dataObjectPathPrefix, someUrl.pathname]);
+        return constructPath([dosDataObjectPathPrefix, someUrl.pathname]);
     }
 }
 
@@ -86,6 +111,14 @@ function dataObjectUriToHttps(dataObjectUri) {
     return output;
 }
 
+// This function counts on the request posing  data as "application/json" content-type.
+// See: https://cloud.google.com/functions/docs/writing/http#parsing_http_requests for more details
+function parseRequest(req) {
+    if (req && req.body) {
+        return req.body.url;
+    }
+}
+
 class FileInfoResponse {
     constructor(contentType, size, timeCreated, updated, md5Hash, bucket, name, gsUri, signedUrl) {
         this.contentType = contentType || '';
@@ -114,7 +147,6 @@ function convertToFileInfoResponse (contentType, size, timeCreated, updated, md5
     );
 }
 
-const samBaseUrl = () => config.samBaseUrl;
 
 class Response {
     constructor(status, data) {
@@ -135,4 +167,4 @@ const promiseHandler = (fn) => (req, res) => {
     return fn(req, res).then(handleValue, handleValue);
 };
 
-module.exports = {dataObjectUriToHttps, convertToFileInfoResponse, samBaseUrl, Response, promiseHandler};
+module.exports = {dataObjectUriToHttps, convertToFileInfoResponse, samBaseUrl, jadeDataRepoUrl, Response, promiseHandler, parseRequest};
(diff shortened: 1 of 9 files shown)


salonishah11:
jenkins retest

common/helpers.js
```diff
@@ -2,24 +2,47 @@ const url = require('url');
 const config = require('../config.json');
 
 const dataGuidsHostPrefix = 'dg.';
-const dataObjectPathPrefix = '/ga4gh/dos/v1/dataobjects/';
+const dosDataObjectPathPrefix = '/ga4gh/dos/v1/dataobjects/';
+const drsDataObjectPathPrefix = '/ga4gh/drs/v1/objects/';
+const jadeDataRepoHostRegex = /jade(-terra)?.datarepo-(dev|prod).broadinstitute.org/;
```
kshakir: I'd make this regex wilder. If when they add an `aou` or a `staging` repo we don't want to have to deploy a new version. perhaps `/jade.*\.datarepo-.*\.broadinstitute\.org/`?

config.json.ctmpl
```diff
@@ -5,6 +5,7 @@
 {
     "dataObjectResolutionHost": {{if eq $runContext "fiab"}}"wb-mock-drs-dev.storage.googleapis.com"{{else if eq $environment "prod"}}"gen3.biodatacatalyst.nhlbi.nih.gov"{{else}}"staging.gen3.biodatacatalyst.nhlbi.nih.gov"{{end}},
     "bondBaseUrl": {{if eq $runContext "fiab"}}"bond-fiab.{{$dnsDomain}}"{{else}}"https://broad-bond-{{$environment}}.appspot.com"{{end}},
-    "samBaseUrl": {{if eq $runContext "fiab"}}"https://sam-fiab.{{$dnsDomain}}"{{else}}"https://sam.dsde-{{$environment}}.broadinstitute.org"{{end}}
+    "samBaseUrl": {{if eq $runContext "fiab"}}"https://sam-fiab.{{$dnsDomain}}"{{else}}"https://sam.dsde-{{$environment}}.broadinstitute.org"{{end}},
+    "jadeDataRepoUrl": {{if eq $environment "prod"}}"jade-terra.datarepo-prod.broadinstitute.org"{{else}}"jade.datarepo-dev.broadinstitute.org"{{end}}
```
kshakir: Is there a reason we need to know the exact jade URLs? `function hasJadeDataRepoHost(someUrl)` can just do a regex match right?

common/helpers.js
```diff
@@ -2,24 +2,47 @@ const url = require('url');
...
     return hasDataGuidsHost(someUrl) || (someUrl.hostname && !someUrl.pathname);
 }
 
 function validateDataObjectUrl(someUrl) {
-    if (hasDataGuidsHost(someUrl) && !someUrl.pathname) {
-        throw new Error(`Data Object URIs with '${dataGuidsHostPrefix}*' host are required to have a path: "${url.format(someUrl)}"`);
+    const hasJDRHost = hasJadeDataRepoHost(someUrl);
+
+    if ((hasDataGuidsHost(someUrl) || hasJDRHost) && !someUrl.pathname) {
+        throw new Error(`Data Object URIs with either '${dataGuidsHostPrefix}*' or '${jadeDataRepoUrl()}' as host are required to have a path: "${url.format(someUrl)}"`);
```
kshakir: Similarly, I would just mention that the url is a type of JDR URL, as there will definitely be more JDRs beyond the `terra` one.

config.json.ctmpl
salonishah11: If we don't do exact host urls, then with right set of permissions one can talk to prod JDR host in qa/dev env no? Do we want to allow that?

kshakir: Good question. Let's tech talk.

My gut reaction it's better to leave it flexible, versus stopping someone from foot-gunning by sending prod requests in dev. Maybe there's a way we can stay flexible and environmentally segregated.

common/helpers.js
salonishah11: I have widened the regex

test/common/helpers.test.js
```diff
@@ -98,4 +140,4 @@ test('dataObjectUriToHttps should throw a Error when passed an invalid uri', (t)
 
 test('samBaseUrl should come from the config json', (t) => {
     t.is(samBaseUrl(), config.samBaseUrl);
-});
+});
```
kshakir, line 152: Farewell sweet newline 💀

config.json.ctmpl
salonishah11: As per the discussion, I have removed the exact matching on host and widened the regex matching. Any data object request which matches the jade data repo regex will be accepted and forwarded to the host mentioned in the request

martha/martha_v3.js
```diff
@@ -0,0 +1,68 @@
...
+        }
+    } else {
+        return Promise.resolve();
+    }
+}
+
+function aggregateResponses(responses) {
+    // Note: for backwards compatibility, we are returning the DRS object with the key, "dos".  If we want to change this, we can add a new api version for Martha_v2.
+    // When/if we change this, we might want to consider replacing "dos" with "data_object" or something like that that is unlikely to change
+    const finalResult = { dos: responses[0] };
```
kshakir, line 25: Yeah, one step at time... can't wait to fix up the replies. Right now `martha_v3` returns very different `response[0]` objects for JDR compared to the other repos.

test/common/helpers.test.js
```diff
@@ -88,6 +88,48 @@ test('should parse "drs://dg." Data Object uri with only a host part with a quer
...
+    t.is(
+        dataObjectUriToHttps('drs://jade-terra.datarepo-prod.broadinstitute.org/anything'),
+        'https://jade-terra.datarepo-prod.broadinstitute.org/ga4gh/drs/v1/objects/anything'
+    );
+});
+
+test('should throw error when given jade data repo host and no path', (t) => {
+    try {
+        dataObjectUriToHttps('drs://jade.datarepo-dev.broadinstitute.org/');
+    } catch(error) {
```
kshakir: What happens in this test when an error is not thrown?

I don't know this test kit yet, but should there be something like `fail("Should have thrown an error but didn't")` before the `catch`?

docker/Dockerfile
```diff
@@ -45,6 +45,7 @@ EXPOSE 8008 8010
 # ENTRYPOINT to be the the `functions start` command.  Deployed functions persist across emulator restarts.
 RUN functions start \
     && functions deploy martha_v2 --trigger-http \
+    && functions deploy martha_v3 --trigger-http \
```
kshakir, line 48: There should probably be an update to the README too, if just a placeholder that says "martha_v3: In early development, don't use"?

martha/martha_v3.js
```diff
@@ -0,0 +1,68 @@
+const {dataObjectUriToHttps, parseRequest} = require('../common/helpers');
+const {bondBaseUrl, determineBondProvider, BondProviders} = require('../common/bond');
+const apiAdapter = require('../common/api_adapter');
+
+
+async function maybeTalkToBond(req, provider = BondProviders.default) {
+    // Currently HCA data access does not require additional credentials.
+    // The HCA checkout buckets allow object read access for GROUP_All_Users@firecloud.org.
+    if (req && req.headers && req.headers.authorization && provider !== BondProviders.HCA) {
```
kshakir, line 9: Needs moar conditionals-- JDR requests are not supposed to contact bond.

test/martha/martha_v3.test.js
```diff
@@ -0,0 +1,154 @@
...
+    await marthaV3(mockRequest({ body: { 'url': 'drs://dg.4503/this_part_can_be_anything' } }), response);
+    const requestedBondUrl = getJsonFromApiStub.secondCall.args[0];
+    const matches = requestedBondUrl.match(bondRegEx);
+    t.truthy(matches, 'Bond URL called does not match Bond URL regular expression');
+    const expectedProvider = 'fence';
+    const actualProvider = matches[2];
+    t.is(actualProvider, expectedProvider);
+});
+
+test.serial('martha_v3 does not call Bond or return SA key when the Data Object URL host endswith ".humancellatlas.org', async (t) => {
```
kshakir, line 146: Related to the prior comment, there should be a test that JDR-ish requests do not talk to bond.

kshakir (commented):
Thanks for talking through the regex change. I had more time post standup for a deeper review, with some questions. Happy to talk "offline" about any of these as well.

martha/martha_v3.js
salonishah11: Yes. Currently whatever JDR returns it will be the value for `dos` key. It will change in WA-160 once there is a standard response schema.

docker/Dockerfile
salonishah11: Oops, I forgot to update README. Will do that.

martha/martha_v3.js
salonishah11: Yes. This is part of [WA-179](https://broadworkbench.atlassian.net/browse/WA-179).

test/martha/martha_v3.test.js
salonishah11: Yes. I have added that test, but as part of [WA-179](https://broadworkbench.atlassian.net/browse/WA-179).

test/common/helpers.test.js
salonishah11: I am not sure either. I will look into it.

salonishah11: There is `t.throws(fn, expectation?, message?)` which can be used. If the test fails to throw an error it will fail the test and return the message passed. I will update the code accordingly.

docker/Dockerfile
salonishah11: done!

kshakir (approved):

common/helpers.js
```diff
@@ -86,6 +98,14 @@ function dataObjectUriToHttps(dataObjectUri) {
     return output;
 }
 
+// This function counts on the request posing  data as "application/json" content-type.
```
rsasch: nit: there are two spaces between "posing" and "data"

rsasch (commented):
Just one question, otherwise 👍: did you intentionally leave duplicate `maybeTalkToBond()` and `aggregateResponses()` in martha_v2.js and martha_v3.js because you figured those would diverge in later PRs?  If not, why not add them helpers.js so they're only in one place?

salonishah11:
>did you intentionally leave duplicate maybeTalkToBond() and aggregateResponses() in martha_v2.js and martha_v3.js because you figured those would diverge in later PRs

@rsasch that's correct. As I mentioned in the PR description, I think they would eventually diverge in [WA-179](https://broadworkbench.atlassian.net/browse/WA-179) and [WA-160](https://broadworkbench.atlassian.net/browse/WA-160).

common/helpers.js
salonishah11: removed the extra space 👍

rsasch (approved):
Ah, ok — sorry I missed that. LGTM

PR #154, opened 2020-04-29
Auth header in request for JDR Integration [WA-179]

salonishah11:
This PR makes auth header being passed to `martha_v3` mandatory. `maybeTalkToBond` has now been refactored into `bond.js`. I initially thought there would be more logic for not talking to Bond in v3 that would keep martha_v3 and martha_v2 code diverged for this function. But it turns out adding 1 additional condition works for both v3 and v2. Integration tests have been added for v3 as part of this PR.

Closes https://broadworkbench.atlassian.net/browse/WA-179.

### DIFF
--- a/common/bond.js
+++ b/common/bond.js
@@ -1,7 +1,7 @@
 const config = require('../config.json');
 const URL = require('url');
 const {hasJadeDataRepoHost} = require('../common/helpers.js');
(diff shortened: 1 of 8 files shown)


salonishah11:
jenkins retest

kshakir (approved):
🙌✅💯

rsasch (approved):

PR #161, opened 2020-05-29
Support for multiple hashes in /martha_v3 [WA-219]

salonishah11:
This PR adds support for new JSON object, “hashes” that contains the keys: hash type, and the corresponding hash in `/martha_v3` response.

Closes https://broadworkbench.atlassian.net/browse/WA-219

### DIFF
--- a/common/helpers.js
+++ b/common/helpers.js
@@ -247,7 +247,7 @@ function getMd5Checksum(checksums) {
  */
 function getHashesMap(checksumArray) {
     /*
-        if there are more than 1 entry for same type of hash, the last hash value in the checksums array for that
+        if there are more than 1 checksum for same type of hash, the last hash value in the checksums array for that
         type will be taken
      */
     return checksumArray.reduce(function(map, obj){
--- a/common/helpers.js
+++ b/common/helpers.js
@@ -118,8 +118,7 @@ class CommonFileInfoResponse {
         bucket,
         name,
         gsUri,
-        googleServiceAccount,
-        signedUrl,
+        googleServiceAccount
     ) {
         this.contentType = contentType || '';
         this.size = size || 0;
@@ -129,7 +128,6 @@ class CommonFileInfoResponse {
         this.name = name || '';
         this.gsUri = gsUri || '';
         this.googleServiceAccount = googleServiceAccount || null;
-        this.signedUrl = signedUrl || '';
     }
 }
 
@@ -146,7 +144,6 @@ class MarthaV3Response extends CommonFileInfoResponse {
         name,
         gsUri,
         googleServiceAccount,
-        signedUrl,
         hashesMap
     ) {
         super(
@@ -157,8 +154,7 @@ class MarthaV3Response extends CommonFileInfoResponse {
             bucket,
             name,
             gsUri,
-            googleServiceAccount,
-            signedUrl
+            googleServiceAccount
         );
         this.hashes = hashesMap || {};
     }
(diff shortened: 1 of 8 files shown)


common/helpers.js
```diff
@@ -257,32 +316,34 @@ function getFileInfoFromDrsResponse(drsResponse, googleServiceAccount) {
...
         mimeType,
         size,
         createdTimeIso,
         updatedTimeIso,
-        md5Checksum,
         bucket,
         name,
         gsUrl,
         googleServiceAccount,
         signedUrl,
```
kshakir: If we're forking the response types, what do you @salonishah11 and other reviewers think of ditching the `signedUrl` for now?

We can always add it back if we re-unify the response types, or actually start returning a signed URL in a future ticket.

My dream of a single response type was to ease transitions for clients, but as the servers are each doing their own thing... Marth'll do what what she has to do.

common/helpers.js
```diff
@@ -200,6 +237,25 @@ function getMd5Checksum(checksums) {
...
+ */
+function getHashesMap(checksumArray) {
+    /*
+        if there are more than 1 checksum for same type of hash, the last hash value in the checksums array for that
+        type will be taken
+     */
+    return checksumArray.reduce(function(map, obj){
+        map[obj.type] = obj.checksum;
+        return map;
+    }, {});
```
kshakir: Beautiful. TIL. 😮👏🥂

Btw, still wish there were _some way_ of adding types to JavaScript/ECMAScript... if only someone out there had come up with something...

common/helpers.js
salonishah11: I agree. Since the responses are already forked, there is no harm in removing `signedUrl` from `/martha_v3` response for now as it is always an empty string.

kshakir (approved):

test/common/helpers.test.js
```diff
@@ -150,3 +150,72 @@ test('dataObjectUriToHttps should throw a Error when passed an invalid uri', (t)
...
+    const expectedChecksumMap = {
+        md5: '336ea55913bc261b72875bd259753046',
+        sha256: 'f76877f8e86ec3932fd2ae04239fbabb8c90199dab0019ae55fa42b31c314c44',
+        crc32c: '8a366443'
+    };
+
+    t.deepEqual(getHashesMap(checksumArray), expectedChecksumMap);
+});
+
+test('getHashesMap should return map for checksums array containing multiple `checksum` for same `type`', (t) => {
```
aednichols: [TOL] I would mildly prefer that we halt and report an error in this case because repeated checksums seems like illegal data

aednichols (approved):

test/common/helpers.test.js
salonishah11: @kshakir What would be your opinion on this?

kshakir: How about-- @salonishah11 if you can complete the change plus update the test in less than 1hr, I say go for @aednichols' suggestion. If you get to 1 hour and you haven't finished, leave it as is?

There absolutely should never be two of the same checksum types returned... so this check should not be required. But there are so many edge cases in the various DOS/DRS versions and implementations that we keep bumping into...

In this case, with the halt-and-report, we'll receive early notification of the issue versus something strange happening for the user and one of us having to track down the mystery.

salonishah11: I have updated the code to throw an error if it encounters a duplicate checksum value for a hash type

kshakir (approved):

aednichols (approved):

PR #162, opened 2020-06-03
Better error response in /martha_v3 [WA-93]

salonishah11:
This PR creates a standard format for error responses and returns better error codes. 

Closes https://broadworkbench.atlassian.net/browse/WA-93

### DIFF
--- a/test/martha/martha_v3.test.js
+++ b/test/martha/martha_v3.test.js
@@ -129,25 +129,37 @@ test.serial('martha_v3 should return 400 if a Data Object without authorization 
     const mockReq = mockRequest({ body: { 'url': 'dos://abc/123' } });
     delete mockReq.headers.authorization;
     await marthaV3(mockReq, response);
+    const result = response.send.lastCall.args[0];
     t.is(response.statusCode, 400);
+    t.is(result.status, 400);
+    t.is(result.response.text, 'Request is invalid. Authorization header is missing.');
 });
 
 test.serial('martha_v3 should return 400 if not given a url', async (t) => {
     const response = mockResponse();
     await marthaV3(mockRequest({ body: { 'uri': 'dos://abc/123' } }), response);
+    const result = response.send.lastCall.args[0];
     t.is(response.statusCode, 400);
+    t.is(result.status, 400);
+    t.is(result.response.text, 'Request is invalid. URL of a DRS object is missing.');
 });
 
 test.serial('martha_v3 should return 400 if no data is posted with the request', async (t) => {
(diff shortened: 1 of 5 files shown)


test/martha/martha_v3.test.js
```diff
@@ -129,25 +129,37 @@ test.serial('martha_v3 should return 400 if a Data Object without authorization
...
 test.serial('martha_v3 should return 400 if given a \'url\' with an invalid value', async (t) => {
     const response = mockResponse();
     await marthaV3(mockRequest({ body: { url: 'Not a valid URI' } }), response);
+    const result = response.send.lastCall.args[0];
     t.is(response.statusCode, 400);
+    t.is(result.status, 400);
+    t.is(result.response.text, 'The specified URL \'Not a valid URI\' is invalid');
 });
 
 test.serial('martha_v3 should return 502 if Data Object resolution fails', async (t) => {
```
kshakir: The `502` in this title no longer matches the `500` changed in the body of the test below. Same goes for "502 if key retrieval from bond fails".

kshakir (changes requested):
Marcus slipped in a [separate feature request into the comment thread of WA-93](https://broadworkbench.atlassian.net/browse/WA-93?focusedCommentId=25842) instead of filing a new ticket. Before we close this ticket, can you either create his ticket for him or implement his request?

It may be as simple as changing [this line](https://github.com/broadinstitute/martha/blob/19edb860261e3281acaba1b24162f52791f604ad/martha/martha_v3.js#L40) to add the [`req.ip`](https://expressjs.com/en/4x/api.html#req.ip).

As for the rest of his request, I believe we [already log](https://github.com/broadinstitute/martha/blob/19edb860261e3281acaba1b24162f52791f604ad/common/api_adapter.js#L62) which URL is returning an error.

One other set of typos found below too.

salonishah11:
@kshakir Oops I somehow forgot to look at the comments in WA-93. I will take a look at your suggestion and add logic to log who is sending this request. For the other request in the same comment (re: detail on which promise failed), I think we had already done this previously, but I will look at the code to make sure we are logging that as well.

Thank you for pointing this out! :)

kshakir (approved):
```
2020-06-04T19:50:45.385Z - info: User function triggered, starting execution
2020-06-04T19:50:45.386Z - info: Received URL 'drs://dg.4503/fa640b0e-9779-452f-99a6-16d833d15bd0' from IP '127.0.0.1'
```

LGTM 👍

rsasch (approved):

PR #168, opened 2020-06-30
Return null instead of empty string in response in /martha_v3 [WA-243]

salonishah11:
Closes https://broadworkbench.atlassian.net/browse/WA-243

### DIFF
--- a/common/helpers.js
+++ b/common/helpers.js
@@ -30,6 +30,10 @@ function validateDataObjectUrl(someUrl) {
     }
 }
 
+function isNullish(value) {
+    return value === null || typeof value === "undefined";
+}
+
 /**
  *  Filter off the null entries first (because `regex.exec(null)` is TRUTHY, because of course it is)
  *  Then run the regex to get the path part without leading or trailing slashes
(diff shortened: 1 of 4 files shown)


kshakir (approved):

common/helpers.js
```diff
@@ -30,6 +30,10 @@ function validateDataObjectUrl(someUrl) {
     }
 }
 
+function isNullish(value) {
+    return value === null || typeof value === "undefined";
```
evanbernstein, line 37: Question, are we checking if something is `null` or `undefined` or are we checking if something is `null` or the string `"undefined"`.  The latter seems surprising (but entirely plausible). If it is the string, do you know what was causing that string?

test/common/helpers.test.js
```diff
@@ -222,3 +222,85 @@ test('getHashesMap should throw error if the checksums array contains duplicate
         'Should have throw error but didnt!'
     );
 });
+
+/**
+ * Test the convertToMarthaV3Response() function
```
evanbernstein, line 227: In addition to these unit tests, what is your thought about adding a test to the overall marthav3 test where something in the DRS response should come back as null in the overall response?

common/helpers.js
salonishah11: The former. It is checking whether it is `null` or `undefined`

test/common/helpers.test.js
salonishah11: The way mocking is set up currently, it is tricky to resolve some fields in the response to null. I will take a look at it to see if it can be done

salonishah11: I was able to add such a test by changing the sandbox setup for only this test

kshakir (approved):

common/helpers.js
evanbernstein: Since you are testing against `undefined` and not the string I don’t think you need that function isNullish. Instead can you try: https://stackoverflow.com/questions/2559318/how-to-check-for-an-undefined-or-null-variable-in-JavaScript I’m not certain though so please try and see what you see. Hope this helps

evanbernstein: Sorry about this. As I mentioned in Slack, I totally missed the `===` and the `typeof`. Thank you for adding the stackoverflow reference so future folks will understand why this was implemented this way!

test/common/helpers.test.js
evanbernstein: Thanks for adding that mocked end-to-end test!

evanbernstein (approved):

PR #173, opened 2020-07-16
martha_v3 endpoint cleanup [WA-268]

rsasch:
- centralize all host-specific code for martha_v3; no more calling to martha_v2-related (outside) functions
- fix inconsistency between `timeCreated` and `timeUpdated` (formerly just `updated`)
- add support in martha_v3 for all hosts/responses covered by martha_v2
- fix martha_v3 tests to reflect different kinds of responses from data-resolving hosts
- get rid of Promises, there is only `await`

Closes https://broadworkbench.atlassian.net/browse/WA-268.


martha/martha_v3.js
```diff
@@ -1,36 +1,159 @@
...
+/** *************************************************************************************************
+ * Response parsers
+ * These are used to take the various kinds of responses that come from the supported data URI
+ * resolvers and generate a standard object from that response.  Like with the URI parsers, the
+ * distinctions are (unfortunately) somewhat arbitrary rules, hence the names are also not
+ * illuminating -- hic sunt dracones.
+ */
+
+function deltaResponseParser (response) {
+    if (response.data_object) {
```
kshakir, line 65: I don't know JavaScript, but based on my experience with typed languages this function should always return or throw _something_.

Perhaps an `else return {};`? Otherwise throw an error

martha/martha_v3.js
```diff
@@ -1,36 +1,159 @@
...
+        return new DrsType(
+            bravoUrlGenerator(parsedUrl),
+            null,
+            echoResponseParser);
+    } else if ((/jade.*\.datarepo-.*\.broadinstitute\.org/).test(parsedUrl.host)) {
+        return new DrsType(
+            charlieUrlGenerator(parsedUrl),
+            null,
+            foxtrotResponseParser);
+    } else if (parsedUrl.host.endsWith('humancellatlas.org')) {
```
kshakir: For "backwards" compatibility with what's in `bond.js`, this should be `endsWith('.humancellatlas.org')` so that `i-am-not-humancellatlas.org` does not match. See `determineBondProvider()` in bond.js.

martha/martha_v3.js
```diff
@@ -1,36 +1,159 @@
...
+}
+
+/** *************************************************************************************************
+ * Here is where all the logic lives that pairs a particular kind of URI with its
+ * resolving-URL-generating parser, what path to use to make a Bond request for an SA (if any), and
+ * a response parser.
+ */
+
+function determineDrsType (parsedUrl) {
+    if (parsedUrl.host.toLowerCase().startsWith('dg.4503')) {
```
kshakir: This does not handle the staging dataguid used for DCP fence. I'm guessing a test is missing. See `STAGING_DATASTAGE_NAMESPACE` and `determineBondProvider()` in bond.js.

martha/martha_v3.js
```diff
@@ -1,36 +1,159 @@
...
+        return new DrsType(
+            bravoUrlGenerator(parsedUrl),
+            `${config.bondBaseUrl}/api/link/v1/fence/serviceaccount/key`,
+            echoResponseParser);
+    } else if (parsedUrl.host.toLowerCase().startsWith('dg.')) {
+        return new DrsType(
+            bravoUrlGenerator(parsedUrl),
+            `${config.bondBaseUrl}/api/link/v1/dcf-fence/serviceaccount/key`,
+            deltaResponseParser);
+    } else if (parsedUrl.host.endsWith('dataguids.org')) {
```
kshakir: This also matches `nottherealdataguids.org`. I don't believe there is an existing function that does this. Should be something like `endsWith('.dataguids.org') or equals('dataguids.org')`.

FYI only: Note also this bug that exists when trying to match host/domain/sub-domain names...: https://news.ycombinator.com/item?id=23479435

martha/martha_v3.js
```diff
@@ -1,36 +1,159 @@
...
+
+function alphaUrlGenerator (parsedUrl) {
+    return url.format({
+        protocol: 'https',
+        hostname: parsedUrl.hostname,
+        port: parsedUrl.port,
+        pathname: `/ga4gh/dos/v1/dataobjects${parsedUrl.pathname}`,
+        search: parsedUrl.search
+    });
+}
```
kshakir: `function dosUrlGenerator (parsedUrl) {`

martha/martha_v3.js
```diff
@@ -1,36 +1,159 @@
...
+    const splitHost = parsedUrl.hostname.split('.');
+    const idPrefix = `${splitHost[0]}.${splitHost[1].toUpperCase()}`;
+    return url.format({
+        protocol: 'https',
+        hostname: config.dataObjectResolutionHost,
+        port: parsedUrl.port,
+        pathname: `/ga4gh/dos/v1/dataobjects/${idPrefix}${parsedUrl.pathname}`,
+        search: parsedUrl.search
+    });
+}
```
kshakir: `function dgDosUrlGenerator (parsedUrl) {`

martha/martha_v3.js
```diff
@@ -1,36 +1,159 @@
...
+
+function charlieUrlGenerator (parsedUrl) {
+    return url.format({
+        protocol: 'https',
+        hostname: parsedUrl.hostname,
+        port: parsedUrl.port,
+        pathname: `/ga4gh/drs/v1/objects${parsedUrl.pathname}`,
+        search: parsedUrl.search
+    });
+}
```
kshakir: `function drsUrlGenerator (parsedUrl) {`

martha/martha_v3.js
```diff
@@ -1,36 +1,159 @@
...
+        return {
+            access_methods: accessMethods,
+            checksums: response.data_object.checksums,
+            created_time: response.data_object.created,
+            mime_type: response.data_object.mimeType || 'application/octet-stream',
+            size: response.data_object.size,
+            updated_time: response.data_object.updated,
+        };
     }
 }
```
kshakir: `function dosResponseParser (response) {`

martha/martha_v3.js
```diff
@@ -1,36 +1,159 @@
...
-        return apiAdapter.getJsonFrom(dataObjectResolutionUrl, auth);
+function baltimoreResponseParser (response) {
+    return {
+        checksums: response.checksums,
+        created_time: response.createdTime,
+        mime_type: response.mimeType || 'application/octet-stream',
+        size: response.size,
+        updated_time: response.updatedTime,
+    };
+}
```
kshakir: Delete this function. A big sign that this function isn't doing anything: this current method throws away both the DOS-style `data_object.urls` and/or the DRS-style `access_methods`, thus is unable to return any URLs to the Martha clients.

You most likely want to call `dosResponseParser` instead, based on the matching call to `dgDosUrlGenerator`.

martha/martha_v3.js
```diff
@@ -1,36 +1,159 @@
...
+
+function casablancaResponseParser (response) {
+    return {
+        access_methods: response.access_methods,
+        checksums: response.checksums,
+        created_time: response.created_time,
+        mime_type: response.mime_type || 'application/octet-stream',
```
kshakir: Remove ` || 'application/octet-stream'`

martha/martha_v3.js
```diff
@@ -1,36 +1,159 @@
...
+        name: response.name,
+        size: response.size,
+        updated_time: response.updated_time,
+    };
+}
```
kshakir: `function drsResponseParser (response) {`

README.md
```diff
@@ -167,7 +167,7 @@ It will always return an object with the same properties:
  contentType:    string,
  size:           int,
  timeCreated:    string [dos objects only],
- updated:        string [usually],
+ timeUpdated:        string [usually],
```
kshakir, line 170: You can always add a field, but you cannot remove `updated` from `fileSummaryV1` unless you are also submitting a _simultaneous_ PR to Firecloud-UI, and I hope your clojure isn't rusty:

- https://github.com/broadinstitute/firecloud-ui/blob/c7bd749eaad0977bc90e4247234f22ddeb8aeb35/src/cljs/main/broadfcui/common/gcs_file_preview.cljs#L108-L112

test/martha/martha_v3.test.js
```diff
@@ -181,7 +227,7 @@ test.serial('martha_v3 should return 400 if given a \'url\' with an invalid valu
     const result = response.send.lastCall.args[0];
     t.is(response.statusCode, 400);
     t.is(result.status, 400);
-    t.is(result.response.text, 'The specified URL \'Not a valid URI\' is invalid');
+    t.is(result.response.text, '"Not%20a%20valid%20URI" is not a properly-formatted URI.');
```
kshakir: ToL: Not a huge deal, but if possible, we should stop encoding the url in error messages like this. Is there a good reason why the value is fed through `url.format()` during `console.error()`?

README.md
```diff
@@ -47,7 +47,7 @@ It will always return an object with the same properties:
  contentType:    string [or null if unknown],
  size:           int [or null if unknown],
  timeCreated:    string [the time created, formatted using ISO 8601, or null if unknown],
- updated:        string [the time updated, formatted using ISO 8601, or null if unknown],
+ timeUpdated:        string [the time updated, formatted using ISO 8601, or null if unknown],
```
kshakir: Remove extra spaces to keep this line spaced with the others.

martha/martha_v3.js
rsasch: This is a distinct function because of the `createdTime` and `updatedTime` parts of the response.

README.md
rsasch: (sigh) ok, added back in for the common and fileSummaryV1, but removed it for the martha v3 response.

martha/martha_v3.js
rsasch: the next check `else if (parsedUrl.host.toLowerCase().startsWith('dg.'))` does

martha/martha_v3.js
rsasch: fixed

martha/martha_v3.js
rsasch: fixed

README.md
rsasch: not needed, I backed out of using `timeUpdated` in the common response class

test/martha/martha_v3.test.js
rsasch: changed it to quote the actual URL submitted

martha/martha_v3.js
rsasch: adding error handling to the parent call and now throw an error

test/martha/martha_v3.test.js
```diff
@@ -84,20 +115,22 @@ const fullExpectedResult = (expectedGoogleServiceAccount) => {
...
+    "data_object": {
+        id: 'v1_abc-123',
+        description: '123 BAM file',
+        name: '123.mapped.abc.bam',
+        created: '2020-04-27T15:56:09.696Z',
+        version: '0',
+        mime_type: 'application/octet-stream',
+        size: 123456
+    }
 };
```
kshakir: Just noticed this-- that's a DOS response but the name of the variable is still `drsObject...` Which is this block supposed to be: `dos` or `drs`?

It may be as simple as renaming this to `dataObjectServiceObjectWithMissingFields`?

test/martha/martha_v3.test.js
```diff
@@ -251,6 +300,7 @@ test.serial('martha_v3 does not call Bond or return SA key when the host url is
 test.serial('martha_v3 returns null for fields missing in drs and bond response', async (t) => {
     // update the stub to return DRS response with missing fields only for this test
     sandbox.restore();
+    getJsonFromApiStub.onFirstCall().resolves(dataObjectServiceObject);
     getJsonFromApiStub = sandbox.stub(apiAdapter, getJsonFromApiMethodName);
     getJsonFromApiStub.onFirstCall().resolves(drsObjectWithMissingFields);
```
kshakir: Ah. I can see here. DOS goes with DOS. DRS always goes with DRS. So this should get renamed.

test/martha/martha_v3.test.js
rsasch: renamed to `dosObjectWithMissingFields`

kshakir (approved):

martha/martha_v3.js
```diff
@@ -40,38 +147,70 @@ function marthaV3Handler(req, res) {
...
+    }
+    const {drsUrl, bondUrl, responseParser} = determineDrsType(parsedUrl);
+    console.log(`Converting DRS URI to HTTPS: ${dataObjectUri} -> ${drsUrl}`);
 
-    console.log(`Received URL '${dataObjectUri}' from IP '${req.ip}'`);
-    let dataObjectResolutionUrl;
+    let response;
     try {
-        dataObjectResolutionUrl = dataObjectUriToHttps(dataObjectUri);
+        response = await apiAdapter.getJsonFrom(drsUrl, auth);
```
kshakir: FYI- Going to submit a PR after this one to add a conditional around this line. We should only send the auth header to an allowed-list of hosts, not all. This line that always-sends-auth is a change from the previous `martha_v3` and `martha_v2` behavior.

rsasch: no need — I added a `sendAuth` property to the DrsType class

rsasch:
tests are currently failing because JDR dev env is down

kshakir (approved):

rsasch:
jenkins retest, if you don't mind

martha/martha_v3.js
```diff
@@ -40,38 +154,74 @@ function marthaV3Handler(req, res) {
...
 
-    return Promise.all([dataObjectPromise, bondPromise])
-        .then((rawResults) => {
-            res.status(200).send(aggregateResponses(rawResults));
-        })
-        .catch((err) => {
-            console.log('Received error while either contacting Bond or resolving drs url.');
+    let bondSA;
+    if (bondUrl && req && req.headers && req.headers.authorization) {
+        try {
```
salonishah11, line 213: TOL: Would it be possible to combine all 3 `try.. catch` blocks in `marthaV3Handler` into single block `try... catch` block? Since most of catch portion is same I think it would be more readable if they are in the same catch block. I know that we would loose the specific error logs pin-pointing the exact location of error. To avoid that, the functions that are called inside the error block could catch these errors locally and add extra information before throwing back the error. So something like
```
let response, drsResponse, bondSA;
try {
  if (sendAuth) {
            response = await apiAdapter.getJsonFrom(drsUrl, auth);
        } else {
            response = await apiAdapter.getJsonFrom(drsUrl);
  }
 drsResponse = responseParser(response);
 if (bondUrl && req && req.headers && req.headers.authorization) {
   bondSA = await apiAdapter.getJsonFrom(bondUrl, auth);
 }
}
catch {
  <insert common error handling here>
}
```

rsasch: I did think about that, but I decided to land on the side of more specific error messages.  I'm open to changing it, though, although I'm not sure how to figure out which call resulted in the error since the majority are calling the same function (`apiAdapter.getJsonFrom()`).

salonishah11 (commented):
Since `martha_v3.js` is no longer using helper function `dataObjectUriToHttps` from `helpers.js`, the functional tests for `dataObjectUriToHttps` ([here](https://github.com/broadinstitute/martha/blob/498bd085c7f3d7e39b3b788f7021205530884acf/test/common/helpers.test.js#L5-L148)) in `helpers.test.js` that check the conversion of drs uri to its https format is no longer checked in martha_v3. Can you add functional tests for `dosFullUrlGenerator`, `dosCompactUrlGenerator` and `drsFullUrlGenerator` that would check whether the conversion from drs uri to it's https for various uris are valid?

martha/martha_v3.js
salonishah11: We could separate out the portion that does drs resolving part and the part that talks to Bond into helper functions. So for example
```
function resolveDrsUri(sendAuth, auth, drsUrl) {
try {
  if (sendAuth) {
            response = await apiAdapter.getJsonFrom(drsUrl, auth);
        } else {
            response = await apiAdapter.getJsonFrom(drsUrl);
  }
 drsResponse = responseParser(response);
}
catch {
 <add additional error information>
}
}

function maybeTalkToBond(bondUrl, req) {
  try{
    if (bondUrl && req && req.headers && req.headers.authorization) {
     bondSA = await apiAdapter.getJsonFrom(bondUrl, auth);
    }
  }
  catch {
     <add additional error information>
  }
}

let response, drsResponse, bondSA;
try {
  drsResponse = resolveDrsUri(sendAuth, auth, drsUrl);
  bondSA = maybeTalkToBond(bondUrl, req);
}
catch {
  <insert common error handling here>
}
```
This again introduces 3 `try... catch` blocks (🤦‍♀️ ) but would make the main `marthaV3Handler` code more readable.
That being said I am not married to the idea of changing separating the 3 `try... catch` blocks. This was just TOL thought 😃

rsasch: Unless you feel strongly about it (or @kshakir weighs in on the other side), I'm content to leave it this way.

kshakir:
> Can you add functional tests for dosFullUrlGenerator, dosCompactUrlGenerator and drsFullUrlGenerator that would check whether the conversion from drs uri to it's https for various uris are valid?

Plus one. For Martha I like both belt (unit tests that don't connect to servers) and suspenders (integration tests, at least for JDR). If the new martha_v3 functions aren't unit tested like the helper.js code then it should be.

> This again introduces 3 try... catch blocks (🤦‍♀️ ) but would make the main marthaV3Handler code more readable.

Minus one. I say leave the main code alone. Martha's various endpoints have been deceptively complex in my humble opinion. I don't mind verbose code in this case. @salonishah11 if you have a suggestion to tighten it up that you'd like to submit as a PR though I'm more than happy to review it.

kshakir (approved):

salonishah11 (approved):

martha/martha_v3.js
```diff
@@ -1,36 +1,156 @@
...
-const { maybeTalkToBond, determineBondProvider, BondProviders } = require('../common/bond');
+const { parseRequest, convertToMarthaV3Response, FailureResponse } = require('../common/helpers');
+const config = require('../config.json');
 const apiAdapter = require('../common/api_adapter');
+const url = require('url');
 
 const BAD_REQUEST_ERROR_CODE = 400;
 const SERVER_ERROR_CODE = 500;
 
+class DrsType {
```
evanbernstein, line 9: Can you tell me about this pattern?  I have not run across it before.  It appears you have created a class with a constructor simply for the sake of creating a return `type` for a function which you immediately destructure. What do you find as the benefit for this pattern?

martha/martha_v3.js
```diff
@@ -1,36 +1,156 @@
...
-function validateRequest(dataObjectUri, auth) {
-    if (!dataObjectUri) {
-        throw new Error('URL of a DRS object is missing.');
-    } else if (!auth) {
-        throw new Error('Authorization header is missing.');
+/** *************************************************************************************************
+ * URI parsers
+ */
+
+function dosFullUrlGenerator (parsedUrl) {
```
evanbernstein, line 22: What is the code documentation philosophy for this team?  I am a bit surprised to see the code having so little code documentation, given that I think the team is a big proponent of documentation in general.  What warrants a function block comment to explain the why and what of the function?

martha/martha_v3.js
```diff
@@ -1,36 +1,156 @@
...
+            false,
+            `${config.bondBaseUrl}/api/link/v1/fence/serviceaccount/key`,
+            dosResponseParser);
+    } else if (parsedUrl.host.toLowerCase().startsWith('dg.')) {
+        return new DrsType(
+            dosCompactUrlGenerator(parsedUrl),
+            false,
+            `${config.bondBaseUrl}/api/link/v1/dcf-fence/serviceaccount/key`,
+            dosResponseParser);
+    } else if (parsedUrl.host.endsWith('.dataguids.org')) {
```
evanbernstein, line 115: I would highly recommend that before the team consider this story complete, that there are mocked end-to-end tests for each branch in this function.  In particular, I do not see any mocked tests for `.dataguids.org` in `martha_v3.test.js` (or elsewhere in the codebase).  I believe we are also missing a test for the `else` branch at the end of this function.  I'm not sure if we think that the current mocked tests that exist in that file cover a significant portion of the control flow that is modified by the behavior that we expect to be different from each server, so I would suggest verifying that too.

martha/martha_v3.js
```diff
@@ -40,38 +160,75 @@ function marthaV3Handler(req, res) {
...
+        console.error(err);
+
+        const errorStatusCode = err.status;
+        if (typeof errorStatusCode === 'undefined') {
+            const failureResponse = new FailureResponse(SERVER_ERROR_CODE, `Received error while parsing response from DRS URL. ${err.message}`);
+            res.status(SERVER_ERROR_CODE).send(failureResponse);
+        } else {
+            res.status(errorStatusCode).send(err);
+        }
+        return;
```
evanbernstein, line 208: This looks an opportunity to write some DRY code. This block of code in the `catch` block is pretty heavily copy and pasted from elsewhere in this function.  This looks like a function that could be refactored out into its own function which takes an error object and a message to do the same work in multiple places.

martha/martha_v3.js
```diff
@@ -1,36 +1,156 @@
...
+    return url.format({
+        protocol: 'https',
+        hostname: parsedUrl.hostname,
+        port: parsedUrl.port,
+        pathname: `/ga4gh/dos/v1/dataobjects${(parsedUrl.pathname) ? parsedUrl.pathname : `/${parsedUrl.hostname}`}`,
+        search: parsedUrl.search
+    });
+}
+
+function dosCompactUrlGenerator (parsedUrl) {
```
evanbernstein, line 32: (very knit-picky detail, but as long as I'm writing other comments) There is a space between the function name and the `(`. This doesn't seem to be the style in the rest of the file but is the style for all newly added functions.  I might suggest consistency over personal preference (going a step further, I'd suggest installing `Prettier` on this codebase so that it auto formats and people never get into style arguments ever).

evanbernstein (commented):
I understand why you wanted to refactor this code to make its decision tree simpler and in one place.  

I know that this PR already merged, and I'm coming in late to look at it.  I have one important comment that I felt was important:  I think that before calling this story done, I'd suggest writing some more mocked end-to-end tests that cover all of the decision tree logic in `determineDrsType`.  Unless I'm missing something, I think some of those branches are untested.  I also have a (potentially) philosophical question about code documentation. And then there are a few less important comments, but as long as I had this one, I thought I'd add them as well.

(None of this is a mandate "from your manager".  This is something for you and the team to think about and discuss)

martha/martha_v3.js
rsasch: This exists elsewhere in the codebase (e.g. the two classes defined here https://github.com/broadinstitute/martha/blob/dev/common/helpers.js#L208); I was following convention.

martha/martha_v3.js
rsasch: I think this would be best addressed at a team meeting.

martha/martha_v3.js
rsasch: I was concerned about containing the scope of this PR because there were previous worries about creep.  Therefore, I sought to only retain existing functionality/tests and not add any.  Any additional tests can have their own issue in Jira.

martha/martha_v3.js
rsasch: This was discussed in earlier comments: see https://github.com/broadinstitute/martha/pull/173#pullrequestreview-450931894 and 
https://github.com/broadinstitute/martha/pull/173#issuecomment-660825418 .

martha/martha_v3.js
rsasch: This seems like it should be a new Jira issue (e.g. installing Prettier).

martha/martha_v3.js
evanbernstein: I saw Saloni's suggestion and Khalid's response.  I would be ok with Saloni's suggestion, but am also fine with your response. What I was suggesting is something slightly different.  Create this helper function:

```javascript
function logError(err, message) {
        console.log(message);
        console.error(err);

        const errorStatusCode = err.status;
        if (typeof errorStatusCode === 'undefined') {
            const failureResponse = new FailureResponse(SERVER_ERROR_CODE, `{$message} ${err.message}`);
            res.status(SERVER_ERROR_CODE).send(failureResponse);
        } else {
            res.status(errorStatusCode).send(err);
        }
        return;
}
```

then use it in the main v3 endpoint:

```javascript
try {
   // first chunk of work
} catch (err) {
	logError(err, "message 1")
}

try {
   // 2nd chunk of work
} catch (err) {
	logError(err, "message 2")
}
```

Some food for thought to get the spirit of what Saloni was requesting while keeping the different error messages that both you and Khalid think are important to keep.

martha/martha_v3.js
evanbernstein: I completely agree! Again, I only added this comment since I was already commenting to say 2 things: 1. I think we should have consistent style, so I would recommend removing those blank spaces next time you touch the code.  And 2. maybe install Prettier in the future (*not* now :) if you think it'd save time (I do).  All of this is extremely low level of importance!

PR #174, opened 2020-07-22
Fix/update README reflecting the code re-factor of martha_v3 [No Jira]

rsasch:


README.md
```diff
@@ -167,7 +164,7 @@ It will always return an object with the same properties:
  contentType:    string,
  size:           int,
  timeCreated:    string [dos objects only],
- timeUpdated:        string [usually],
+ updated:        string [usually],
```
aednichols: I know this is a pre-existing condition, but what does "usually" mean?

aednichols (approved):

salonishah11 (approved):

README.md
kshakir: Can come back as `null` also. I'm not an expert in JSON, but I believe than null is a different type than a string.

rsasch: To put a finer point on it, Martha is at the mercy of whatever it gets back from the URI resolver, so there's no guarantee that it won't be `null`.

aednichols: Cool. This is very much a seagull/out of scope request but can we say
```suggestion
 updated:        string [string by design, resolver sometimes returns null]
```
to capture this conversation?

PR #183, opened 2020-08-07
Mock test Martha v3 [WA-287]

aednichols:
Used example responses from [this document](https://docs.google.com/document/d/1Wf4enSGOEXD5_AE-uzLoYqjIp5MnePbZ6kYTVFp1WoM/edit#) to test how Martha handles a variety of response formats.

Some of these look broken to me (lots of `null`s) but seeing as the goal was to test existing behavior, not change it, I am submitting as-is. It's also possible I am grossly misunderstanding how Martha works, so this could also be an opportunity to set me straight.


test/martha/martha_v3.test.js
```diff
@@ -279,30 +536,86 @@ test.serial('martha_v3 calls bond Bond with the "fence" provider when the Data O
...
+    t.true(getJsonFromApiStub.calledOnce); // Bond was not called to get SA key
+    t.deepEqual(Object.assign({}, result), kidsFirstDrsMarthaResult(null));
+    t.falsy(result.googleServiceAccount);
+    t.is(response.statusCode, 200);
+});
+
+test.serial('martha_v3 parses HCA response correctly', async (t) => {
+    getJsonFromApiStub.onFirstCall().resolves(hcaDosResponse);
+    const response = mockResponse();
+    await marthaV3(mockRequest({ body: { 'url': 'dos://jade.datarepo-dev.broadinstitute.org/abc' } }), response);
```
salonishah11: The host name in the request url should end with HCA, for example `dos://someservice.humancellatlas.org/abc`. I think this is the reason the response has most of the values as null. Since the request has JDR host, it tries to parse the response using DRS v1 schema instead of the DOS schema resulting in null values for almost everything. 

Similarly the test for Anvil, Kids first, BDC and Gen3 should also have their respective hostnames in the request

salonishah11 (changes requested):

test/martha/martha_v3.test.js
evanbernstein: As Saloni said, hopefully this explains why you have the comment above `hcaDosMarthaResult`: `Quite a grim result, but could be working as expected?`

Also, I think we might want to add another part to each of these tests.  `determineDrsType` determines what we should be sending to our 3rd parties, what the URL and parameters look like.  As a part of testing the boundaries, we may want to test that as a part of these tests.  We should probably test the arguments that are being sent to  `getJsonFromApiStub.onFirstCall()`.

test/martha/martha_v3.test.js
```diff
@@ -160,25 +417,25 @@ test.serial.afterEach(() => {
...
     await marthaV3(mockRequest({ body: { 'url': 'dos://abc/123' } }), response);
     const result = response.send.lastCall.args[0];
-    t.deepEqual(Object.assign({}, result), fullExpectedResult(googleSAKeyObject));
+    t.deepEqual(Object.assign({}, result), jadeDosMarthaResult(googleSAKeyObject));
     t.is(response.statusCode, 200);
 });
 
 test.serial('martha_v3 resolves a valid DRS-style url', async (t) => {
-    getJsonFromApiStub.onFirstCall().resolves(dataObjectServiceObject);
+    getJsonFromApiStub.onFirstCall().resolves(jadeDosResponse);
```
evanbernstein: Is this really correct that a DRS url resolves to DOS?  If so, maybe add a comment to this test explaining why this odd result is correct?

test/martha/martha_v3.test.js
```diff
@@ -279,30 +536,86 @@ test.serial('martha_v3 calls bond Bond with the "fence" provider when the Data O
...
+    await marthaV3(mockRequest({ body: { 'url': 'drs://jade.datarepo-dev.broadinstitute.org/abc' } }), response);
+    const result = response.send.lastCall.args[0];
+    t.true(getJsonFromApiStub.calledOnce); // Bond was not called to get SA key
+    t.deepEqual(Object.assign({}, result), jadeDosMarthaResult(null));
+    t.falsy(result.googleServiceAccount);
+    t.is(response.statusCode, 200);
+});
+
+test.serial('martha_v3 parses Gen3 CRDC response correctly', async (t) => {
+    getJsonFromApiStub.onFirstCall().resolves(gen3CrdcResponse);
```
evanbernstein, line 219: As Saloni said, but for clarity, this test should call gen3, not JDR. I'm commenting on this one specifically since the line of code is outside of the diff.

evanbernstein (changes requested):

test/martha/martha_v3.test.js
```diff
@@ -279,30 +536,86 @@ test.serial('martha_v3 calls bond Bond with the "fence" provider when the Data O
...
+    await marthaV3(mockRequest({ body: { 'url': 'dos://jade.datarepo-dev.broadinstitute.org/abc' } }), response);
+    const result = response.send.lastCall.args[0];
+    t.true(getJsonFromApiStub.calledOnce); // Bond was not called to get SA key
+    t.deepEqual(Object.assign({}, result), kidsFirstDrsMarthaResult(null));
+    t.falsy(result.googleServiceAccount);
+    t.is(response.statusCode, 200);
+});
+
+test.serial('martha_v3 parses HCA response correctly', async (t) => {
+    getJsonFromApiStub.onFirstCall().resolves(hcaDosResponse);
```
evanbernstein, line 263: Per my suggestion below, I looked how to do this in Sinon, if you are interested in pursuing it. Per https://sinonjs.org/releases/v9.0.2/stubs/ :
```suggestion
    getJsonFromApiStub.withArgs([fill in args here]).onFirstCall().resolves(hcaDosResponse);
```
You just need to fill in what the params should be to `getJsonFrom` when called from the url `dos://jade.datarepo-dev.broadinstitute.org/abc`.

test/martha/martha_v3.test.js
aednichols: Nice, when updated to the URL Saloni suggested the response is reasonable. This tells me the test is checking something valuable, because there is evidently important special handling on the URL.

Next question (also noted in a code comment), why is `size` a string in HCA but an int elsewhere?

[I recognize that the goal of this story is to capture existing behavior, but I think apparent inconsistency warrants examination]

salonishah11: Huh, that's interesting. What happens when you keep the size as an Int in the response [here](https://github.com/broadinstitute/martha/blob/51df8c8837020b096bc40cfc9370542d29b10b0a/test/martha/martha_v3.test.js#L365)? Does the test fail? The size in response should be Int as per [martha_v3's response schema](https://github.com/broadinstitute/martha#martha-v3).

salonishah11: I think it's because of [this](https://github.com/broadinstitute/martha/blob/630f1b7af769f1db08d78d6798c3d6ed0fad3d3b/martha/martha_v3.js#L76) line of code. So it looks like if the `size` in the result that you get back from external DRS server is an `Int`, `size` in the final martha_v3 response will be an `Int`. Same goes for it being a `String`.

aednichols: Yes, the test fails when the types do not match.

I think it's not a good design to vary the output type based on the input, would you agree if I modified the code to always cast to int?

salonishah11: Yes I agree that we should not vary the output based on input. I think you can change the code since you found a bug.

aednichols: Cool, done.

aednichols:
I think I've addressed everything except nice-to-haves

test/martha/martha_v3.test.js
```diff
@@ -31,350 +48,6 @@ const mockResponse = () => {
...
-        name: "blobs/44d320c9606e32d6df04d8a4023e6474efaa2f99de436e07f7241942972d5703.c3349268e528c844c528a427aa13034e72b7b39d.16c68f2306435b7cf990153b37adeb20-134.64d51664",
-        gsUri: "gs://org-hca-dss-checkout-prod/blobs/44d320c9606e32d6df04d8a4023e6474efaa2f99de436e07f7241942972d5703.c3349268e528c844c528a427aa13034e72b7b39d.16c68f2306435b7cf990153b37adeb20-134.64d51664",
-        googleServiceAccount: expectedGoogleServiceAccount,
-        hashes: {
-            sha256: '44d320c9606e32d6df04d8a4023e6474efaa2f99de436e07f7241942972d5703'
-        }
-    };
-};
-
 const dosObjectWithMissingFields = {
```
evanbernstein: Why remove all the other testing constants to `_martha_v3_resources.js` but not this and the next? I'm just curious. (Any of the non-shared ones, I could also argue could just be in the test function itself.)  I'm not trying to change your opinion.

test/martha/martha_v3.test.js
```diff
@@ -553,17 +226,17 @@ test.serial('martha_v3 does not call Bond or return SA key when the host url is
...
 });
 
 test.serial('martha_v3 parses Gen3 CRDC response correctly', async (t) => {
     getJsonFromApiStub.onFirstCall().resolves(gen3CrdcResponse);
     const response = mockResponse();
-    await marthaV3(mockRequest({ body: { 'url': 'drs://jade.datarepo-dev.broadinstitute.org/abc' } }), response);
+    await marthaV3(mockRequest({ body: { 'url': 'drs://nci-crdc.datacommons.io/asdfasdfasdf' } }), response);
     const result = response.send.lastCall.args[0];
-    t.true(getJsonFromApiStub.calledOnce); // Bond was not called to get SA key
+    // t.true(getJsonFromApiStub.calledOnce); // Bond was not called to get SA key
```
evanbernstein: Should this extra commented line be removed?

evanbernstein (approved):
This looks good to me (obviously assuming you resolve the CI issues).  I wouldn't mind creating a follow-up ticket, if you and the team agree, on testing the parameters to the 3rd party URLs as I had suggested as a bonus in this PR.

test/martha/martha_v3.test.js
```diff
@@ -160,25 +91,25 @@ test.serial.afterEach(() => {
...
     await marthaV3(mockRequest({ body: { 'url': 'dos://abc/123' } }), response);
     const result = response.send.lastCall.args[0];
-    t.deepEqual(Object.assign({}, result), fullExpectedResult(googleSAKeyObject));
+    t.deepEqual(Object.assign({}, result), sampleDosMarthaResult(googleSAKeyObject));
     t.is(response.statusCode, 200);
 });
 
 test.serial('martha_v3 resolves a valid DRS-style url', async (t) => {
-    getJsonFromApiStub.onFirstCall().resolves(dataObjectServiceObject);
+    getJsonFromApiStub.onFirstCall().resolves(sampleDosResponse);
```
evanbernstein, line 87: This prior comment got lost in a code change: Is this really correct that a DRS url resolves to DOS? If so, maybe add a comment to this test explaining why this odd result is correct?

test/martha/martha_v3.test.js
aednichols: No reason, it's a lot shorter and my pattern-matcher missed it

test/martha/martha_v3.test.js
aednichols: Good catch, it should stay in and be uncommented

test/martha/martha_v3.test.js
aednichols: I'm not sure, this particular test is pre-existing (I just renamed the variable)

Will find out & fix if appropriate

salonishah11 (approved):

PR #186, opened 2020-09-14
The AnVIL updates and test updates GAWB-4053 WA-319

kshakir:
- `martha_v3` mock tests ensure the correct Bond link has been contacted
- Added The AnVIL to the list of DOS/DRS servers that need to talk to Bond in `martha_v3`
- Stopping enumerating and comparing servers that don't talk to Bond, just use undefined or `null`
- Add `getSignedUrlV1` tests to ensure non-Bond DOS/DRS servers query their SA from SAM
- Update the Bond prefix in the committed `config.json` to send `https` instead of `http`
- Use `https` to talk to FiaB Bond instead of relying on the server to force the redirection
- Updated test regex to account for `https` in the Bond configs
- Update tests to remove static link so Bond host can be changed using `BOND_BASE_URL`
- When functions only require part of an object only pass that part
- More use of `&&` shortcutting instead of using `if`
- Add guess-comment as to where the `martha_v3` fields `bucket` and `name` originated
- Consolidate and add comment on converting `size` to a `Number()`
- `DrsType` fields now contain dg.* expansion and the `provider` to contact on Bond
- Composing instead of duplicating URL generator logic across DOS/DRS vs. Full/Compact
- Consolidated response parsers into one DOS vs. DRS response parser
- Preserve `dg.aBcD` casing by switching from the deprecated `url.parse()` to `new URL()`
- Updated tests to use `Error` text returned by `new URL()`
- Added more mock tests including a `https://dataguids.org` DOS resolution test
- In mocks of real servers use real URIs so folks can try them out if they have access
- Don't pass `expectedGoogleServiceAccount` to mock providers that will never use the value
- Do pass `expectedGoogleServiceAccount` to mock providers that should return the SA
- Swapped in a new mock HCA response and added a `curl` reference as to where it came from
- Use object spread instead of `Object.assign()` and validate using `eslint`
- Updated README.md based on URLs discovered in other prior updates to the mock tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ It will always return an object with the same properties:
  name:                  string [resolver sometimes returns null],
  gsUri:                 string [resolver sometimes returns null],
  googleServiceAccount:  object [null unless the DOS url belongs to a Bond supported host],
+ fileName:              string [resolver sometimes returns null],
  hashes:                object [contains the hashes type and their checksum value; if unknown, it returns null]
 ```
 
@@ -65,6 +66,7 @@ Example response for /martha_v3:
     "name": "dd3c716a-852f-4d74-9073-9920e835ec8a/f3b148ac-1802-4acc-a0b9-610ea266fb61",
     "gsUri": "gs://my-bucket/dd3c716a-852f-4d74-9073-9920e835ec8a/f3b148ac-1802-4acc-a0b9-610ea266fb61",
     "googleServiceAccount": null,
+    "fileName": "hello.txt",
     "hashes": {
         "md5": "336ea55913bc261b72875bd259753046",
         "sha256": "f76877f8e86ec3932fd2ae04239fbabb8c90199dab0019ae55fa42b31c314c44",
@@ -72,6 +74,18 @@ Example response for /martha_v3:
     }
 }
 ```
+
+The fields are:
+- `gsUri`: The full Google Cloud Storage URI/URL/path to the blob storing the data
+- `bucket`: The [bucket name](https://cloud.google.com/storage/docs/key-terms#bucket-names) part of the `gsUri`
+- `name`: The [object name](https://cloud.google.com/storage/docs/key-terms#object-names) part of the `gsUri`
+- `fileName`: The file name for the blob found at `gsUri`
+- `contentType`: The type of data stored in the blob found at `gsUri`
+- `size`: The size of the blob found at `gsUri`
+- `hashes`: The various hash types and values for the blob found at `gsUri`
+- `timeCreated`: The time of creation for the data found at `gsUri`
+- `timeUpdated`: The time of last update for the data found at `gsUri`
+- `googleServiceAccount`: An optional service account that should be used to access the `gsUri`
 
 **NOTE:**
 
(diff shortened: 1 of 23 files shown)


kshakir:
Draft PRs that will use these changes:
- https://github.com/DataBiosphere/terra-ui/pull/2233 (PR targets https://github.com/DataBiosphere/terra-ui/pull/2223, that's *33* -> *23*)
- https://github.com/broadinstitute/cromwell/pull/5850 (Centaur won't pass tests until these changes to `martha_v3` are deployed)

README.md
```diff
@@ -65,6 +66,7 @@ Example response for /martha_v3:
     "name": "dd3c716a-852f-4d74-9073-9920e835ec8a/f3b148ac-1802-4acc-a0b9-610ea266fb61",
     "gsUri": "gs://my-bucket/dd3c716a-852f-4d74-9073-9920e835ec8a/f3b148ac-1802-4acc-a0b9-610ea266fb61",
```
aednichols: I don't understand how this standard was usable at all before the addition of `fileName`

If someone gives me a URI like `drs://test/some-uuid` and I dereference it, I get the GCS directory path... but no usable file paths. Am I supposed to `gsutil ls` in my code to get the filenames? Seems like a big omission

This is a rounadabout way of saying, good thing you implemented this, it seems essential

kshakir: JDR was the first server that was a) using this optional field and b) that we noticed. The DOS/DRS servers currently being used in production return GCS paths that end in usable file names so no one has complained yet.

aednichols: Ohhhh whaaaaaat? How is it possible to write code that consumes these URIs without knowing whether it points to a file or not? Clients will have to write sensor code to parse the string and guess whether it's a file or not

kshakir: Yeah, we can discuss within WA and will have to extend discussions to the GATK, but this is not new, just not common in bioinformatics IMHO.

There's been a constant battle of "please use the file extension to define the file type" for decades. Still historically [MIME types](https://en.wikipedia.org/wiki/Media_type), [MacOS resource forks](https://en.wikipedia.org/wiki/Resource_fork), and now DOS/DRS have had separate concepts of file names vs. file types.

common/bond.js
```diff
@@ -23,35 +23,41 @@ const bondBaseUrl = () => config.bondBaseUrl;
...
+        return BondProviders.ANVIL;
+    }
+
+    if (url.host.endsWith('.humancellatlas.org')) {
+        return;
+    }
+
+    if (hasJadeDataRepoHost(url)) {
+        return;
+    }
```
rsasch: why not combine this into one `if` statement?
```
if (url.host.endsWith('.humancellatlas.org') || hasJadeDataRepoHost(url)) {
        return;
}

File: common/helpers.js [hunk 3]
```diff
@@ -338,19 +352,43 @@ function convertToMarthaV3Response(drsResponse, googleSA) {
...
+        size: maybeNumberSize,
         updated_time: updatedTime,
+        name: maybeFileName,
     } = drsResponse;
 
     // Some (but not all!) DRS servers return time without a timezone (see example responses in `_martha_v3_resources.js`)
     // Instead of letting JS assume that a timezoneless time is local TZ, explicitly assign it to be UTC
     const createdTimeIso = createdTime ? moment.utc(createdTime).toISOString() : null;
     const updatedTimeIso = updatedTime ? moment.utc(updatedTime).toISOString() : null;
+
```
rsasch, line 369: nit: empty line seems unnecessary

martha/martha_v3.js
```diff
@@ -100,62 +94,71 @@ function drsResponseParser (response) {
...
-        return new DrsType(
-            dosCompactUrlGenerator(parsedUrl),
-            false,
-            `${config.bondBaseUrl}/api/link/v1/dcf-fence/serviceaccount/key`,
-            dosResponseParser);
-    } else if (parsedUrl.host.endsWith('.dataguids.org')) {
+    const host = parsedUrl.host.toLowerCase();
+
+    // First handle servers that we know about...
+
```
rsasch, line 103: nit: empty line seems unnecessary

test/common/bond_test.js
```diff
@@ -32,7 +36,7 @@ test('determineBondProvider should be "dcf-fence" if the URL host is "dg.foo"',
 });
 
 test('determineBondProvider should be "HCA" if the URL host ends with ".humancellatlas.org"', (t) => {
```
rsasch: please update the description of the test to reflect the new logic
e.g.
```suggestion
test('determineBondProvider should not return a provider if the URL host ends with ".humancellatlas.org"', (t) => {
```

test/common/bond_test.js
```diff
@@ -45,5 +49,5 @@ test('determineBondProvider should return the default BondProvider if the URL ho
 });
 
 test('determineBondProvider should return the "jade-data-repo" as the provider for JDR host"', (t) => {
```
rsasch: please update the description of the test to reflect the new logic

test/martha/martha_v3.test.js
```diff
@@ -281,28 +363,38 @@ test.serial('martha_v3 returns null for fields missing in drs and bond response'
...
 
+function generateUrl(testUrl) {
+    const parsedUrl = new URL(testUrl);
+    return determineDrsType(parsedUrl).urlGenerator(parsedUrl);
+}
 
 /**
  * determineDrsType(uri) -> drsUrl Scenario 1: data objects uri with non-dg host and path
  */
 test('determineDrsType should parse dos:// Data Object uri', (t) => {
```
rsasch, line 380: please update the description of the test to reflect the new logic
e.g.
```suggestion
test('generateUrl should return correct URL from dos:// Data Object URI', (t) => {
```

rinse, repeat for the rest of the `generateURL` tests

rsasch (changes requested):
small changes and nits

common/bond.js
kshakir: I realize [guard clauses](https://www.google.com/search?q=javascript+guard+clause) are usually aimed at nested conditionals, but I also consider lots of `|| && ()` mixed together in a conditional soup harder to read than multiple lines.

rsasch (approved):

common/bond.js
rsasch: TOL If you think it's likely that more `||`s or `&&`s will be added, maybe a helper function instead?

common/bond.js
```diff
@@ -23,35 +23,41 @@ const bondBaseUrl = () => config.bondBaseUrl;
...
-    } else {
-        return BondProviders.default;
     }
+
+    if (ANVIL_NAMESPACE === url.hostname.toLowerCase()) {
+        return BondProviders.ANVIL;
+    }
+
+    if (url.host.endsWith('.humancellatlas.org')) {
+        return;
```
aednichols, line 39: I don't understand how you can return a const in one place and nothing in another. Doesn't this make it hard for the caller to use this function?

kshakir: This isn't returning nothing. It's a compact way of [returning `undefined`](https://codeburst.io/javascript-what-is-the-return-statement-97d8b11a1a0c). Yay for multiple empty types in JavaScript.

I'm still learning ES6 and the latest coding styles. [Here's a four year old discussion on `null` vs. `undefined`.](https://stackoverflow.com/questions/37980559/is-it-better-to-return-undefined-or-null-from-a-javascript-function) I would love to know best practices, and hopefully an `eslint` rule to enforce them so I don't forget.

kshakir:
Update: removed WA-344 implementation in the latest commit. Updated the PR title. Discussions are ongoing with Jade, ingesters, etc. to figure out where on earth to get the file names from.

martha/martha_v3.js
```diff
@@ -6,91 +6,85 @@ const url = require('url');
 const BAD_REQUEST_ERROR_CODE = 400;
 const SERVER_ERROR_CODE = 500;
 
+const BOND_LINK_NONE = null; // Used for servers that should NOT contact bond
+const BOND_LINK_DCF_FENCE = 'dcf-fence'; // The default when we don't recognize the server
+const BOND_LINK_FENCE = 'fence';
+const BOND_LINK_ANVIL = 'anvil';
```
salonishah11: Why not use BondProviders from [here](https://github.com/broadinstitute/martha/blob/3a1acc3f9b68fcfe3c118d16ce87eba3822128af/common/bond.js#L6-L13)?

test/common/bond_test.js
```diff
@@ -44,6 +48,6 @@ test('determineBondProvider should return the default BondProvider if the URL ho
     t.is(determineBondProvider('drs://some-host/anything'), BondProviders.default);
 });
 
-test('determineBondProvider should return the "jade-data-repo" as the provider for JDR host"', (t) => {
-    t.is(determineBondProvider('drs://jade.datarepo-dev.broadinstitute.org/identifier'), BondProviders.JADE_DATA_REPO);
+test('determineBondProvider should not return a provider if the URL host is JDR"', (t) => {
+    t.falsy(determineBondProvider('drs://jade.datarepo-dev.broadinstitute.org/identifier'));
 });
```
salonishah11, line 53: Can we add more tests with Anvil and Kids First DRS uri here to check they return expected Bond providers?

martha/martha_v3.js
kshakir: Unlike the other endpoints, `martha_v3` reduced its dependencies and reimplements the common logic in `bond.js`. I'm happy to consider changing this if we discuss.

salonishah11 (approved):

kshakir:
I noticed some strangeness in the mock response from the AnVIL stored in our code: https://github.com/broadinstitute/martha/blob/1ea49d64b77b4a1eed18f5241c453a3b1984e516/test/martha/_martha_v3_resources.js#L227-L230

I put in another guard clause and tests to make sure if someone in prod is using URIs with the full host name like that then the correct Bond provider is used.

README.md
```diff
@@ -92,48 +103,67 @@ Martha's `martha_v3` implementation translates requests-to and responses-from th
     - Martha Testing: 🤖 Continuous Automated
     - Returns Bond SA: No
     - Requires OAuth for metadata: 🔐 Yes
-- ❌ [DataGuids.org](https://dataguids.org/) (ex: any drs://dg.* other than drs://dg.4503, and not drs://dataguids.org)
+    - Example: `drs://jade.datarepo-dev.broadinstitute.org/v1_0c86170e-312d-4b39-a0a4-2a2bfaa24c7a_c0e40912-8b14-43f6-9a2f-b278144d0060`
+- ❌ [DataGuids.org](https://dataguids.org/)
+    (ex: any drs://dg.* other than drs://dg.4503, drs://dg.712C, drs://dg.ANV0, and not drs://dataguids.org)
```
rsasch: nit: since you're explicitly adding examples, I think it makes sense to remove the "ex" from "(ex: ...)".
(this applies to all the hosts)

common/helpers.js
rsasch: still true

martha/martha_v3.js
```diff
@@ -6,156 +6,207 @@ const url = require('url');
...
-function dosCompactUrlGenerator (parsedUrl) {
-    const splitHost = parsedUrl.hostname.split('.');
-    let idPrefix;
-    if (parsedUrl.pathname) {
-        idPrefix = `${splitHost[0]}.${splitHost[1].toUpperCase()}`;
-    } else {
-        idPrefix = `${splitHost[0]}.${splitHost[1]}`;
-    }
-
+const compactUrlGenerator = (dataGuidExpansion) => function (parsedUrl, protocolPrefix) {
```
rsasch, line 49: why is this function declaration in this format instead of
```suggestion
function compactUrlGenerator (parsedUrl, dataGuidExpansion, protocolPrefix) {
```
like `httpsUrlGenerator` ?

martha/martha_v3.js
```diff
@@ -6,156 +6,207 @@ const url = require('url');
...
-            dosResponseParser);
-    } else if (parsedUrl.host.endsWith('.dataguids.org')) {
+            DG_EXPANSION_NONE,
+            PROTOCOL_PREFIX_DOS,
+            AUTH_SKIPPED,
+            BOND_PROVIDER_FENCE,
+        );
+    }
+
+    // Compact the AnVIL
```
rsasch: nit: why the "the"?

martha/martha_v3.js
```diff
@@ -6,156 +6,207 @@ const url = require('url');
...
-            dosResponseParser);
-    } else if ((/jade.*\.datarepo-.*\.broadinstitute\.org/).test(parsedUrl.host)) {
+            DG_EXPANSION_THE_ANVIL,
+            PROTOCOL_PREFIX_DOS,
+            AUTH_SKIPPED,
+            BOND_PROVIDER_ANVIL,
+        );
+    }
+
+    // Full the AnVIL
```
rsasch: nit: why the "the"?

martha/martha_v3.js
```diff
@@ -6,156 +6,207 @@ const url = require('url');
...
+            DG_EXPANSION_NONE,
+            PROTOCOL_PREFIX_DRS,
+            AUTH_REQUIRED,
+            BOND_PROVIDER_NONE,
+        );
+    }
+
+    // Assume BDC compact but with dcf-fence
+    // If we don't recognize the dg.* assume like martha_v2 that everyone else
+    // speaks DOS, doesn't require auth, and uses dcf-fence
```
rsasch: nit: this comment is hard to read because it contains a sentence fragment all on one line that is capitalized but doesn't have a "." at the end and then a full sentence that is capitalized, split between two lines, and also doesn't have a "." at the end

salonishah11 (approved):

rsasch (approved):
small changes and nits, otherwise 👍

README.md
kshakir: There are negative examples in this list, the "other than". Off the top of my head I don't know how to put those in negative examples as bullet points, but I also hope `dos://dataguids.org/1234` goes away so we can just remove this whole entry from the readme.

martha/martha_v3.js
kshakir: Because of how it's called. I (semi-)[curried the function](https://medium.com/javascript-scene/curry-and-function-composition-2c208d774983) and then only partially applied it before as its used below.

martha/martha_v3.js
kshakir: 🤷 Go ask the AnVIL: https://gen3.theanvil.io/

martha/martha_v3.js
kshakir: 🤷 Go ask the AnVIL: https://gen3.theanvil.io/

PR #193, opened 2020-10-19
Updated Bond Providers for DOS/DRS servers BA-6643

kshakir:


test/common/bond_test.js
```diff
@@ -67,14 +67,14 @@ test('determineBondProvider should return the AnVIL BondProvider if the URL host
 test('determineBondProvider should return the default BondProvider if the URL host is Kids First', (t) => {
```
gpolumbo-broad:
```suggestion
test('determineBondProvider should return the dcf-fence BondProvider if the URL host is Kids First', (t) => {
```

test/common/bond_test.js
```diff
@@ -86,7 +86,7 @@ test(
                 'drs://drs.dev.singlecell.gi.ucsc.edu' +
                 '/bee7a822-ea28-4374-8e18-8b9941392723?version=2019-05-15T205839.080730Z'
             ),
-            BondProviders.default,
+            BondProviders.DCF_FENCE,
```
gpolumbo-broad, line 89: Again, tweak the test name to explicitly state `dcf-fence` instead of `default` Bond Provider.

gpolumbo-broad (changes requested):

gpolumbo-broad (approved):

kshakir:
Will rebase this onto #195 after that is reviewed and merged.

grsterin (approved):

.gitignore
```diff
@@ -1,7 +1,12 @@
 /node_modules
 .idea
-**/*.private.env.json
-**/deploy_account.json
-**/*.pem
```
breilly2, line 5: Isn't the `**/*` syntax for catching files in any subdirectory? I worry that, without it, it will only catch files in the base directory. I'd have to read the gitignore docs; I don't use it frequently enough to remember.

kshakir: > Isn't the **/* syntax for catching files in any subdirectory?

Sorta.

> If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular .gitignore file itself. Otherwise the pattern may also match at any level below the .gitignore level.

via: https://git-scm.com/docs/gitignore#_pattern_format

So these are all equivalent:
- `*.pem`
- `**/*.pem`
- `**/**/*.pem`
- `**/**/**/*.pem`
- etc.

I went for the first option.

PR #195, opened 2020-10-19
Upgrade dependencies and ignore manually upgraded "unfixed" trivy CVE BA-6652

kshakir:


docker/Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM node:12.16.1
+# NOTE: If editing this file then check if cleanup of .trivyignore is needed also!
+FROM node:12-buster
```
breilly2: We should specifically ask for `12.19.0-buster` to build reproducibility.

kshakir: We actually don't care about the version of Node. Eventually Martha's various endpoints ends up in GCF where we have no control over the minor version. Most likely (hopefully! 🤞) Google is constantly upgrading the GCFs to the latest Node point releases with security patches, etc.

So, these images aren't used to \*run\* the eventual deployments. They're used a) to emulate in a FiaB and b) later \*to\* deploy the GCF. (This image is quite strange and contains a lot.) Still, for consistency with the other Firecloud/Terra images that are used in production I'll freeze the minor version.

breilly2:
Is the alpine-based image vulnerable? Would martha work on alpine?

kshakir:
No clue if we can deploy/emulate-for-FiaB on alpine. Never tried it. 🤷

docker/Dockerfile
```diff
@@ -37,3 +38,14 @@ EXPOSE 8010
 
 ENTRYPOINT ["npm"]
 CMD ["start"]
+
+# Need to update for Trivy found CVE-2019-20367 in libbsd0 0.9.1-2
+# Feel free to remove after upgrading the docker image
```
breilly2:
```suggestion
# Feel free to remove after upgrading the docker image to a debian release newer than buster
```

package.json
```diff
@@ -20,17 +20,17 @@
...
-    "ava": "^3.12.1",
-    "eslint": "^7.9.0",
+    "ava": "^3.13.0",
+    "eslint": "^7.11.0",
     "minimist": "^1.2.5",
     "nyc": "^15.1.0",
-    "sinon": "^9.0.3",
-    "supertest": "^4.0.2"
+    "sinon": "^9.2.0",
+    "supertest": "^5.0.0"
```
breilly2, line 34: Not much in the way of changelog info for these, but they appear to be small changes and they're dev dependencies anyway.

breilly2 (commented):
Overall looks good, though I didn't test it manually (not sure how) so I'm relying on the test suite for that. I mainly just want to have the docker image version fully specified before giving a thumb.

docker/Dockerfile
breilly2: Oh dear. Well, in that case, maybe we should be non-specific. I misunderstood the purpose of this Dockerfile.
When deploying the cloud function, do we at least control the major version?

kshakir: We do control the major deploy version, here:

https://github.com/broadinstitute/martha/blob/9277913f57c6d5b6da5b3e1debd9ff0520c830df/deploy.sh#L111-L117

As for how-the-sausage-is-made, you'll notice the above deploys actually on 10. Here's another inconsistency where we code against 12, which matches what we currently emulate on:

- https://github.com/broadinstitute/martha/blob/9277913f57c6d5b6da5b3e1debd9ff0520c830df/.node-version#L1

For the lurkers, relevant links (constantly updating):
- https://nodejs.org/en/about/releases/
- https://cloud.google.com/functions/docs/concepts/nodejs-runtime

kshakir:
Btw,

> though I didn't test it manually (not sure how)

The [docs are short](https://github.com/broadinstitute/martha/tree/9277913f57c6d5b6da5b3e1debd9ff0520c830df#run-tests) and correct. It will also print out a coverage report showing you what lines were not tested.

mcovarr (approved):

docker/Dockerfile
breilly2: Ugh, I hate that it's so inconsistent, but that's not the purpose of this PR so I'll leave it alone for now.

breilly2:
> Btw,
> 
> > though I didn't test it manually (not sure how)
> 
> The [docs are short](https://github.com/broadinstitute/martha/tree/9277913f57c6d5b6da5b3e1debd9ff0520c830df#run-tests) and correct. It will also print out a coverage report showing you what lines were not tested.

I meant more real-world testing. I trust that the automated tests are running... automatically... but that doesn't prove that the change is valid. We do have some test coverage elsewhere, though, and I judge the risk to be low based on the changes being made.

breilly2 (approved):
👍  Looks good.

kshakir:
> I meant more real-world testing

No. We do not have real-world testing.

Only recently has @ ndbolliger shepherded proof-of-concept "RAS" based testing you may have heard of. Someday we'll get to implementing that. Until then, we cross our fingers and test after we deploy to dev while we try to release to production: https://github.com/broadinstitute/martha/blob/9277913f57c6d5b6da5b3e1debd9ff0520c830df/docs/release-checklist.md

PR #196, opened 2020-10-26
Remove code for martha_v2 endpoint [BW-355]

rsasch:
Closes https://broadworkbench.atlassian.net/browse/BW-355


kshakir (approved):

salonishah11 (changes requested):
Can you please also update **Martha_v1** & **Martha_v2** sections README to reflect that martha_v2 has been removed and instead martha_v3 should be used?

salonishah11 (approved):

rsasch:
jenkins retest

rsasch:
jenkins retest

kv076 (commented):
wondering if we should use a URL shortener instead? Are Broad.io URLs public?(I assume yes)

rsasch:
> wondering if we should use a URL shortener instead? Are Broad.io URLs public?(I assume yes)

The URL will be rendered as a click-able link, so I'm not sure why a URL shortener would be necessary.

PR #199, opened 2020-11-09
Fix for resolving dg.712C vs dg.712c BT-24

kshakir:


test/martha/martha_v3.test.js
```diff
@@ -525,6 +525,31 @@ test.serial('martha_v3 parses BDC staging response correctly', async (t) => {
     t.is(actualProvider, expectedProvider);
 });
 
+test.serial('martha_v3 parses uppercase BDC staging response correctly', async (t) => {
```
breilly2: Except for `712c` -> `712C`, is this exactly the same as the test above? I don't think we'd lose any meaningful coverage if we test with just `712C`.

kshakir: Happy to switch to only testing `712C` capitalized 👍

breilly2 (approved):
👍  Looks good. I'm interested in reducing the complexity of the test cases if possible, but not with strong enough conviction to block this PR.

breilly2 (approved):

salonishah11 (approved):

grsterin (approved):

PR #202, opened 2021-01-13
Remove Trivy scan [BT-86]

dinvlad:
This change removes Trivy scan from CI, since the Docker image is only used in FiaBs and is not needed to be scanned for compliance. This should reduce the associated noise for you 😉


breilly2 (approved):
👍  Looks good. Thanks!

grsterin (approved):

PR #210, opened 2021-04-07
[BT-186] Martha returns signed URLs for open-access BDC data

kshakir:


martha/martha_v3.js
```diff
@@ -398,31 +424,66 @@ async function marthaV3Handler(req, res) {
...
     } catch (error) {
         logAndSendServerError(res, error, 'Received error while parsing response from DRS URL.');
         return;
     }
 
+    // do more stuff here, i.e. get signed URL
+    // first, fetch access token from bond
+    // TODO: allow this to be not done based on requested data
+    let accessToken;
+    if (bondAccessTokenUrl && accessId) {
```
kshakir: - [x] Replace the comments with something more descriptive
- [x] Don't access this `if` if the client hasn't requested the `accessUrl`

martha/martha_v3.js
```diff
@@ -398,31 +424,66 @@ async function marthaV3Handler(req, res) {
...
+            accessUrl = await apiAdapter.getJsonFrom(httpsAccessUrl, accessTokenAuth);
+        } catch (error) {
+            logAndSendServerError(res, error, 'Received error contacting DRS provider.');
+            return;
+        }
+    }
+
     let bondSA;
-    if (bondUrl && overlapFields(requestedFields, BOND_FIELDS)) {
+    if (bondSAKeyUrl && overlapFields(requestedFields, BOND_FIELDS)) {
```
kshakir: `BOND_FIELDS` should be renamed. It is now the fields that imply one should contact bond for the SA key

martha/martha_v3.js
```diff
@@ -76,11 +80,24 @@ const DG_COMPACT_KIDS_FIRST = 'dg.f82a1a';
...
+    constructor(urlParts, protocolPrefix, sendAuth, bondProvider, accessMethodType) {
         this.urlParts = urlParts;
         this.protocolPrefix = protocolPrefix;
         this.sendAuth = sendAuth;
         this.bondProvider = bondProvider;
+        this.accessMethodType = accessMethodType;
+    }
+}
+
+/* Returns undefined if the matching access method does not have an access_id. */
```
mcovarr: ... or if the `accessMethodType` is falsy or the `drsResponse` is falsy

README.md
```diff
@@ -67,6 +68,12 @@ Example response for /martha_v3:
     "gsUri": "gs://my-bucket/dd3c716a-852f-4d74-9073-9920e835ec8a/f3b148ac-1802-4acc-a0b9-610ea266fb61",
     "googleServiceAccount": null,
     "fileName": "hello.txt",
+    "accessUrl": {
+      "url": "https://storage.example.com/f3b148ac-1802-4acc-a0b9-610ea266fb61?sig=ABC",
+      "headers": {
+        "Authorization": "Basic Z2E0Z2g6ZHJz"
+      }
+    },
```
mcovarr, line 76: nice!

common/helpers.js
```diff
@@ -266,14 +268,27 @@ const promiseHandler = (fn) => (req, res) => {
 /**
  * Extracts the bucket and path from a Google Cloud Storage URL
  *
- * @param uri The GCS url
+ * @param {?string} uri The GCS url
```
mcovarr: uri and url being used interchangeably

kshakir: _Matches the code._ /s 🤦

Mini rant: GCS technically [uses a URI](https://cloud.google.com/bigquery-transfer/docs/cloud-storage-transfer-overview#google-cloud-storage-uri) while consistently a lot of code around the various DSP repos "wings it" by using classes like the [WHATWG URL](https://nodejs.org/api/url.html#url_the_whatwg_url_api) to parse the values.

And then of course these are not `URL` objects but `string`...

martha/martha_v3.js
```diff
@@ -406,23 +486,67 @@ async function marthaV3Handler(req, res) {
         return;
     }
 
-    let bondSA;
-    if (bondUrl && overlapFields(requestedFields, BOND_FIELDS)) {
+    // Try to retrieve the file name from the initial DRS response
+    let fileName = getDrsFileName(drsResponse);
+
+    // If we do NOT have the fileName yet, we will try to get that field from the accessUrl
+    const accessIdFields = fileName ? MARTHA_V3_ACCESS_ID_FIELDS : MARTHA_V3_ACCESS_ID_FN_FIELDS;
```
kshakir: To discuss: It should be possible to get the `fileName` only from the metadata without requesting the signed URL.

martha/martha_v3.js
```diff
@@ -174,7 +232,7 @@ function getHttpsUrlParts(url) {
     // The many different ways a DOS/DRS may be "compact", in the order that the should be tried
     const cibRegExps = [
         // Non-W3C CIB DOS/DRS URIs, where the `dg.abcd` appears more than once
-        /(?:dos|drs):\/\/(?<host>dg\.[0-9a-z-]+)(?<separator>:)(?:\k<host>)\/(?<suffix>[^?]*)(?<query>\?(.*))?/i,
+        /(?:dos|drs):\/\/(?<host>dg\.[0-9a-z-]+)(?<separator>:)\k<host>\/(?<suffix>[^?]*)(?<query>\?(.*))?/i,
```
kshakir, line 233: FYI IntelliJ suggested this edit.

deploy.sh
```diff
@@ -108,13 +108,13 @@ docker run \
...
+      gcloud beta functions deploy martha_v2 --trigger-http --source=. --runtime nodejs12 \\
         --allow-unauthenticated --project ${DEPLOY_PROJECT_NAME} &&
-      gcloud beta functions deploy martha_v3 --trigger-http --source=. --runtime nodejs10 \\
+      gcloud beta functions deploy martha_v3 --trigger-http --source=. --runtime nodejs12 \\
         --allow-unauthenticated --project ${DEPLOY_PROJECT_NAME} &&
-      gcloud beta functions deploy fileSummaryV1 --trigger-http --source=. --runtime nodejs10 \\
+      gcloud beta functions deploy fileSummaryV1 --trigger-http --source=. --runtime nodejs12 \\
         --allow-unauthenticated --project ${DEPLOY_PROJECT_NAME} &&
-      gcloud beta functions deploy getSignedUrlV1 --trigger-http --source=. --runtime nodejs10 \\
+      gcloud beta functions deploy getSignedUrlV1 --trigger-http --source=. --runtime nodejs12 \\
```
kshakir, line 117: FYI: This aligns "real" GCF with the nodejs version used by the emulator run inside FiaB containers: https://github.com/broadinstitute/martha/blob/3.5.0/docker/Dockerfile#L2

test/martha/martha_v3.test.js
```diff
@@ -63,36 +69,49 @@ const mockResponse = () => {
...
+    token: 'my-fake-token',
+    expires_at: 'NEVER'
+};
+
+function mockGcsAccessUrl(gsUrlString) {
+    const gsUrl = new URL(gsUrlString);
+    return { url: `https://storage.googleapis.com/${gsUrl.hostname}${gsUrl.pathname}?sig=ABC` };
+}
+
 const bondRegEx = /^https:\/\/([^/]+)\/api\/link\/v1\/([a-z-]+)\/serviceaccount\/key$/;
```
mcovarr: plz bondServiceAccountRegEx

test/martha/martha_v3.test.js
```diff
@@ -526,22 +780,27 @@ test.serial('martha_v3 parses BDC staging response correctly', async (t) => {
...
+    t.deepEqual({ ...result }, anvilDrsMarthaResult(googleSAKeyObject, drsAccessUrlResponse));
     t.is(
-        getJsonFromApiStub.firstCall.args[0],
+        getJsonFromApiStub.getCall(0).args[0],
         `https://${config.HOST_THE_ANVIL_STAGING}/ga4gh/drs/v1/objects/dg.ANV0/00008531-03d7-418c-b3d3-b7b22b5381a0`,
     );
-    t.falsy(getJsonFromApiStub.firstCall.args[1]); // no auth passed
-    const requestedBondUrl = getJsonFromApiStub.secondCall.args[0];
+    t.falsy(getJsonFromApiStub.getCall(0).args[1]); // no auth passed
+    const requestedBondUrl = getJsonFromApiStub.getCall(3).args[0];
```
mcovarr: hey what happened to calls 1 and 2

test/martha/martha_v3.test.js
```diff
@@ -155,53 +176,244 @@ test.serial('martha_v3 resolves successfully and ignores extra data submitted be
...
+    const requestedBondAccessToken = getJsonFromApiStub.getCall(1).args[0];
+    const accessTokenMatches = requestedBondAccessToken.match(bondAccessTokenRegEx);
+    t.truthy(accessTokenMatches, 'Bond URL called does not match Bond URL regular expression');
+    const expectedAccessTokenProvider = 'fence';
+    const actualAccessTokenProvider = accessTokenMatches[2];
+    t.is(actualAccessTokenProvider, expectedAccessTokenProvider);
+
+    t.is(
+        getJsonFromApiStub.getCall(2).args[0],
+        'https://staging.gen3.biodatacatalyst.nhlbi.nih.gov/ga4gh/drs/v1/objects' +
```
kshakir: Replace these hostnames with constants as is used in the other matches.

martha/martha_v3.js
kshakir: Decision: Return `null` and do not access the signed URL to sleuth out the file name

martha/martha_v3.js
breilly2: Makes sense. It would throw off some match groups if referenced by position, but we're naming them all so it should be okay.

test/martha/martha_v3.test.js
```diff
@@ -354,25 +351,8 @@ test.serial('martha_v3 calls the correct endpoints when only the fileName is req
...
-    t.is(actualAccessTokenProvider, expectedAccessTokenProvider);
-
-    t.is(
-        getJsonFromApiStub.getCall(2).args[0],
-        `https://${config.HOST_BIODATA_CATALYST_STAGING}/ga4gh/drs/v1/objects` +
-        '/dg.712C/fa640b0e-9779-452f-99a6-16d833d15bd0/access/gs',
-    );
-    t.is(getJsonFromApiStub.getCall(2).args[1], `Bearer ${bondAccessTokenResponse.token}`);
+    sinon.assert.callCount(getJsonFromApiStub, 1); // Bond was not called to retrieve the googleServiceAccount
+    t.deepEqual({ ...result }, { fileName: null });
```
mcovarr, line 326: TOL is `{ ...result }` not the same as `result`?

kshakir: In this case, probably. There were (are?) similar test cases where `deepEqual` tested the [equality of the wrapper](https://github.com/avajs/ava/blob/main/docs/03-assertions.md#deepequalvalue-expected-message).

For consistency and because a lot of these lines are later copy-pasted I used the same "erase the wrapper type" syntax here.

mcovarr (approved):

test/martha/martha_v3.test.js
```diff
@@ -107,99 +135,272 @@ test.serial('martha_v3 resolves a valid DOS-style url', async (t) => {
...
+    const result = response.send.lastCall.args[0];
+    sinon.assert.callCount(getJsonFromApiStub, 1); // File name available from the metadata
+    t.deepEqual(
+        { ...result },
+        mask(bdcDrsMarthaResult(googleSAKeyObject, drsAccessUrlResponse), 'fileName'),
+    );
+});
+
+test.serial('martha_v3 calls the correct endpoints when only the fileName is requested and the metadata contains a gs url', async (t) => {
+    const drsResponse = bdcDrsResponseCustom({
```
breilly2: This results in an unmodified `bdcDrsResponse`. As a result, I think this behavior is already covered by `martha_v3 parses BDC response correctly`.

kshakir: Yes, true today, and happy to remove this as a separate test for now. If/when BDC changes what fields get populated we can change both `bdcDrsResponse` and `bdcDrsResponseCustom` at that time.

test/martha/martha_v3.test.js
```diff
@@ -351,321 +552,449 @@ test.serial('martha_v3 should return the underlying status if Data Object resolu
...
 test.serial('martha_v3 parses Gen3 CRDC response correctly', async (t) => {
-    getJsonFromApiStub.onFirstCall().resolves(gen3CrdcResponse);
+    const drsAccessUrlResponse = mockGcsAccessUrl(gen3CrdcResponse.access_methods[0].access_url.url);
+    getJsonFromApiStub.onCall(0).resolves(gen3CrdcResponse);
+    getJsonFromApiStub.onCall(1).resolves(bondAccessTokenResponse);
+    getJsonFromApiStub.onCall(2).resolves(drsAccessUrlResponse);
+    getJsonFromApiStub.onCall(3).resolves(googleSAKeyObject);
     const response = mockResponse();
     await marthaV3(
         mockRequest({ body: { 'url': `dos://${config.HOST_CRDC_STAGING}/206dfaa6-bcf1-4bc9-b2d0-77179f0f48fc` } }),
```
breilly2, line 609: Should this be `drs://`?

kshakir: Despite attempts to remove DOS from the internet, it did exist and still does.

breilly2 (approved):
👍

PR #211, opened 2021-04-09
Refactor martha_v3 handler into smaller functions

kshakir:


martha/martha_v3.js
```diff
@@ -468,22 +460,45 @@ async function marthaV3Handler(req, res) {
         `with auth required '${sendAuth}' and bond provider '${bondProvider}'`
     );
 
+    Object.assign(opt, {
+        requestedFields,
+        httpsMetadataUrl,
+        auth,
```
mcovarr: It looks like `auth` and `requestedFields` were not changed since their creation by destructuring from `opt` (they are `const`s), so is this just restructuring the same values back into `opt`?

martha/martha_v3.js
```diff
@@ -426,37 +427,28 @@ function overlapFields(requestedFields, serviceFields) {
...
-function sendResponse(res, requestedFields, drsResponse, fileName, bondProvider, bondSA, accessUrl) {
-    const fullResponse =
-        requestedFields.length
-            ? convertToMarthaV3Response(drsResponse, fileName, bondProvider, bondSA, accessUrl)
-            : {};
-    const partialResponse = mask(fullResponse, requestedFields.join(","));
-
-    res.status(200).send(partialResponse);
-}
+function parseOptions(opt) {
```
mcovarr: nit: some of these "options" are not optional, e.g. `url` and `auth`. Perhaps `parseParameters(params)` or something like that?

martha/martha_v3.js
```diff
@@ -468,22 +460,45 @@ async function marthaV3Handler(req, res) {
...
+        drsType,
+        sendAuth,
+        bondProvider,
+        accessMethodType,
+        bondSAKeyUrl,
+        bondAccessTokenUrl,
+    });
+}
+
+async function retrieveData(opt) {
```
mcovarr: more non-options in this `opt`

martha/martha_v3.js
```diff
@@ -534,12 +546,70 @@ async function marthaV3Handler(req, res) {
...
-    sendResponse(res, requestedFields, drsResponse, fileName, bondProvider, bondSA, accessUrl);
+    Object.assign(opt, {
+        drsResponse,
+        fileName,
+        accessUrl,
+        bondSA,
+    });
+}
+
+function makeResponse(opt) {
```
mcovarr: here too

martha/martha_v3.js
mcovarr: Unrelated to your changes but it might be helpful to have a link to Michael's Lucid Chart sequence diagram(s) here.

martha/martha_v3.js
```diff
@@ -426,37 +427,28 @@ function overlapFields(requestedFields, serviceFields) {
...
-    const {authorization: auth, 'user-agent': userAgent} = req.headers;
-    console.log(`Received URL '${url}' from agent '${userAgent}' on IP '${req.ip}'`);
+    console.log(`Received URL '${url}' from agent '${userAgent}' on IP '${ip}'`);
 
     try {
         validateRequest(url, auth, requestedFields);
     } catch (error) {
-        logAndSendBadRequest(res, error);
-        return;
+        throw new BadRequestError(error);
```
breilly2: `validateRequest` could throw `BadRequestError`s, and then this function doesn't need to `try`/`catch`.

mcovarr (approved):

martha/martha_v3.js
```diff
@@ -534,12 +552,70 @@ async function marthaV3Handler(req, res) {
...
+
+        res.status(200).send(params.partialResponse);
+    } catch (error) {
+        if (error instanceof BadRequestError) {
+            logAndSendBadRequest(res, error.cause);
+        } else if (error instanceof RemoteServerError) {
+            logAndSendServerError(res, error.cause, error.description);
+        } else {
+            console.error(`Uncaught error: ${error}`);
+            throw error;
```
breilly2, line 609: Since we chatted about this a little in the mob, I remembered that letting an exception escape from a cloud function causes GCP to destroy that instance instead of letting it continue handling requests. Since we don't know what went wrong by the time we get here, killing the instance is probably a good idea. 👍

martha/martha_v3.js
```diff
@@ -468,22 +460,51 @@ async function marthaV3Handler(req, res) {
...
         } catch (error) {
-            logAndSendServerError(res, error, 'Received error while resolving DRS URL.');
-            return;
+            throw new RemoteServerError(error, 'Received error while resolving DRS URL.');
         }
     }
 
     let drsResponse;
     try {
         drsResponse = responseParser(response);
```
breilly2, line 493: Can this even throw an error? I didn't notice anything in `responseParser`.

breilly2 (approved):
👍   Looks good. Only very minor questions/nits.

martha/martha_v3.js
breilly2: Nevermind. There's code in `responseParser` that deals with converting DOS responses to DRS responses that I suspect could throw errors because I don't have enough knowledge of the DOS spec. I retract my implied suggestion to remove the `try`/`catch` from here.

PR #216, opened 2021-04-18
Tests blocked by GCS access url support: Check accessUrl for requester pays when passed a googleBillingProject

kshakir:


common/api_adapter.js
```diff
@@ -1,14 +1,21 @@
...
 
 const MAX_RETRY_ATTEMPTS = 5;
 const INITIAL_BACKOFF_DELAY = 1000;
 const BACKOFF_FACTOR = 2;
 
+const GCS_REQUESTER_PAYS_MESSAGE = 'Bucket is a requester pays bucket but no user project provided.';
+
+const BAD_REQUEST_CODE = 400;
+const FORBIDDEN_CODE = 403;
 const TOO_MANY_REQUESTS_CODE = 429;
```
mikebaumann, line 12: The HTTP code currently returned by Gen3 in the case of rate limiting is 503.
They may use 429 in the future, and their tech recommended  we check for both 503 and 429, in this Data Biosphere Slack post: https://databiosphere.slack.com/archives/CSN5HM1EV/p1579722138021900

PR #217, opened 2021-04-21
Return access method type plus some test polish

kshakir:


kshakir:
Going with changing Martha to try-and-silence accessUrl generation failures.

PR #219, opened 2021-04-21
Return null if accessUrl generation times out BT-259

kshakir:


martha/martha_v3.js
```diff
@@ -527,7 +532,26 @@ async function retrieveFromServers(params) {
...
+            result in the remaining short GCF lifecycle.
+
+            NOTE: The call to getJsonFrom will not be canceled, and will keep pointlessly WA-90 retrying. But the
+            upstream client to martha_v3 might get a response if there's time left.
+             */
+            // noinspection JSCheckFunctionSignatures
+            const accessUrlAttempt = await Promise.race([
+                apiAdapter.getJsonFrom(httpsAccessUrl, accessTokenAuth),
+                sleep(REQUEST_TIMEOUT_MILLIS, REQUEST_TIMEOUT_RESULT),
+            ]);
```
kshakir, line 561: To do: if there's _any_ error returned, even not a timeout, then keep going and return `accessUrl: null`

common/helpers.js
```diff
@@ -1,6 +1,6 @@
 const url = require('url');
 const config = require('../common/config');
-
+const sleep = require('util').promisify(setTimeout);
```
breilly2, line 3: I believe this will be fine. https://nodejs.org/api/util.html#util_util_promisify_original mentions the pitfall with `this`, but that isn't a problem here.

martha/martha_v3.js
breilly2: Agreed. Once we have that, I believe this is good to go.

kshakir: Updated and added a test.

kshakir:
This is going to be redone. Closing so this doesn't confuse the Jira board but feel free to re-use the testing logic.

PR #212, opened 2021-04-12
Kids First DRS Signed URLs [BT-172] [BT-218]

mcovarr:

PR #220, opened 2021-04-21
Temporarily turn off BDC signed URLs [BT-236]

mcovarr:


martha/martha_v3.js
```diff
@@ -70,6 +70,7 @@ const BOND_PROVIDER_FENCE = 'fence';
 const BOND_PROVIDER_ANVIL = 'anvil';
 
 const ACCESS_METHOD_TYPE_NONE = null;
+// eslint-disable-next-line no-unused-vars
 const ACCESS_METHOD_TYPE_GCS = 'gs';
```
kshakir: ToL: One can also comment out the variable?

test/martha/martha_v3.test.js
```diff
@@ -81,6 +81,7 @@ function mockGcsAccessUrl(gsUrlString) {
 
 const bondSAKeyUrlRegEx = /^https:\/\/([^/]+)\/api\/link\/v1\/([a-z-]+)\/serviceaccount\/key$/;
 
+// eslint-disable-next-line no-unused-vars
 const bondAccessTokenUrlRegEx = /^https:\/\/([^/]+)\/api\/link\/v1\/([a-z-]+)\/accesstoken$/;
```
kshakir: Same ToL: could be commented out

test/martha/martha_v3.test.js
```diff
@@ -653,15 +621,14 @@ test.serial('martha_v3 parses a Gen3 CRDC CIB URI response correctly', async (t)
...
 test.serial('martha_v3 parses BDC response correctly', async (t) => {
     const drsAccessUrlResponse = mockGcsAccessUrl(bdcDrsResponse.access_methods[0].access_url.url);
     // Rotated once from Michael Baumann's sequence diagram so that call 0 is the second call down from the top,
     // call 3 is the top call.
     // https://lucid.app/lucidchart/428a0bdd-a884-4fc7-9a49-7bf300ef6777/edit?shared=true&page=0_0#
     getJsonFromApiStub.onCall(0).resolves(bdcDrsResponse);
-    getJsonFromApiStub.onCall(1).resolves(bondAccessTokenResponse);
-    getJsonFromApiStub.onCall(2).resolves(drsAccessUrlResponse);
-    getJsonFromApiStub.onCall(3).resolves(googleSAKeyObject);
+    getJsonFromApiStub.onCall(1).resolves(googleSAKeyObject);
```
kshakir: FYI: There's no more "call 3", and I just noticed the sequence diagram doesn't have a link.

test/martha/martha_v3.test.js
```diff
@@ -670,12 +637,12 @@ test.serial('martha_v3 parses BDC response correctly', async (t) => {
     );
     t.is(response.statusCode, 200);
     const result = response.send.lastCall.args[0];
-    // Bond was called twice to get SA key and access token.
-    // DRS server was called twice to get DRS response and signed URL.
-    // 2 + 2 = 4.
-    sinon.assert.callCount(getJsonFromApiStub, 4);
+    // Bond was called once to get SA key.
+    // DRS server was called once to get DRS response.
+    // 1 + 1 = 2.
```
kshakir: Now that you're a certified Martha expert, what are your thoughts on these types of comments existing _only_ in this one test in the middle of the test suite?

test/martha/martha_v3.test.js
```diff
@@ -670,12 +637,12 @@ test.serial('martha_v3 parses BDC response correctly', async (t) => {
...
+    sinon.assert.callCount(getJsonFromApiStub, 2);
 
-    t.deepEqual({ ...result }, bdcDrsMarthaResult(googleSAKeyObject, drsAccessUrlResponse));
+    t.deepEqual({ ...result }, {...bdcDrsMarthaResult(googleSAKeyObject, drsAccessUrlResponse), accessUrl: null});
```
kshakir: Would this be the same as:
```javascript
t.deepEqual({ ...result }, bdcDrsMarthaResult(googleSAKeyObject, null));
```

test/martha/martha_v3.test.js
```diff
@@ -684,65 +651,36 @@ test.serial('martha_v3 parses BDC response correctly', async (t) => {
...
     await marthaV3(
         mockRequest({ body: { 'url': 'drs://dg.712C/fc046e84-6cf9-43a3-99cc-ffa2964b88cb' } }),
         response,
     );
     t.is(response.statusCode, 200);
     const result = response.send.lastCall.args[0];
-    sinon.assert.callCount(getJsonFromApiStub, 4);
-    t.deepEqual({ ...result }, bdcDrsMarthaResult(googleSAKeyObject, drsAccessUrlResponse));
+    sinon.assert.callCount(getJsonFromApiStub, 2);
+    t.deepEqual({ ...result }, {...bdcDrsMarthaResult(googleSAKeyObject, drsAccessUrlResponse), accessUrl: null});
```
kshakir: Same replacing `drsAccessUrlResponse` with null.

test/martha/martha_v3.test.js
```diff
@@ -911,7 +849,8 @@ test.serial('martha_v3 should return 500 if Data Object parsing fails', async (t
     );
 });
 
-test.serial('martha_v3 should return 500 if access method parsing fails', async (t) => {
+/* BT-236 Skip this as it fails in weird ways (not in 'access method parsing') with signed URLs turned off. */
+test.skip('martha_v3 should return 500 if access method parsing fails', async (t) => {
```
kshakir, line 845: This test _really_ needs to be redone at some point. IMHO there's no reason to expect `drsResponse.access_methods is not iterable` to ever occur, so I think the error should not get special error handling.

For now, I'm fine leaving this and will continue to try and get this fixed in one of our upcoming PRs.

test/martha/martha_v3.test.js
```diff
@@ -933,7 +872,9 @@ test.serial('martha_v3 should return 500 if access method parsing fails', async
     );
 });
 
-test.serial('martha_v3 should return 500 on exception trying to get access token from Bond', async (t) => {
+/* BT-236 Skip testing access token fetch failure since that is not something this code even attempts with BDC
+ * signed URLs turned off. */
+test.skip('martha_v3 should return 500 on exception trying to get access token from Bond', async (t) => {
```
kshakir, line 869: I don't like leaving commented out code like this (search `ignore` in the Cromwell repo) but hopefully we'll be back soon 🤞

martha/martha_v3.js
mcovarr: or delete it, whichever my reviewers like best 🙂

test/martha/martha_v3.test.js
mcovarr: agreed this one was particularly gross

test/martha/martha_v3.test.js
mcovarr: yeah I'm not imagining this as a "TODO" that never gets "TODONE" but as a PR that will be completely reverted in a week or two.

test/martha/martha_v3.test.js
mcovarr: apparently so

test/martha/martha_v3.test.js
mcovarr: this can probably go but since the whole PR will be reverted they may sneak back in 🙂

test/martha/martha_v3.test.js
mcovarr: that link appears to be working for me...

kshakir: 🤦

kshakir (approved):
Fine with the way it is

test/martha/martha_v3.test.js
```diff
@@ -332,10 +316,7 @@ test.serial('martha_v3 calls returns the DRS name field for a file name even whe
         access_url: { url: null },
         access_id: bdcDrsResponse.access_methods[0].access_id,
     });
-    const drsAccessUrlResponse = mockGcsAccessUrl('gs://mybucket/from_access_url.txt');
```
breilly2, line 335: This test is `martha_v3 calls returns the DRS name field for a file name even when it differs from the access url`. We can't really test that if we don't fetch an access URL. Maybe we can switch this test over to a hypothetical mock Kids First call?

test/martha/martha_v3.test.js
```diff
@@ -232,11 +233,9 @@ test.serial('martha_v3 calls the correct endpoints the googleServiceAccount is r
     t.is(actualSAKeyProvider, expectedSAKeyProvider);
 });
 
+// BT-236 temporarily cut access token and access endpoint out of the flow
```
breilly2, line 235: Not wrong, but maybe it would be more useful to switch this test over to use a DrsType and DRS response that will lead to an access URL being requested.

breilly2 (commented):
A couple of comments. I think we need to modify a test or two to maintain current coverage.

test/martha/martha_v3.test.js
mcovarr: There are Kids First tests including the access URL flow added in #212 which is only blocked on this PR

test/martha/martha_v3.test.js
mcovarr: ~I believe this should also be covered with #212~ hmm "should be" but isn't... I will try adding that to #212

test/martha/martha_v3.test.js
breilly2: Okay, as long as we reconcile this once both PRs are merged. Right now, this test is "martha_v3 doesn't call /access if the DRS response doesn't have an access_id, even if accessUrl is the only field requested from Martha".

test/martha/martha_v3.test.js
breilly2: Okay, I understand how these 2 PRs are co-dependent. Not ideal, but it should work out once both are merged, and we can clean up later if needed.

breilly2 (approved):
👍   Looks good based on additional test coverage that is coming in #212.

mcovarr:
jenkins retest

PR #222, opened 2021-05-17
proposal of changes to how we use sinon

breilly2:
(includes turning on BDC signed URLs for demonstration purposes)

I can walk through this in person, but here's what I'm thinking we can do with these tests to make them a little more flexible (but still equal coverage) and readable.

### DIFF
--- a/test/martha/martha_v3.test.js
+++ b/test/martha/martha_v3.test.js
@@ -49,11 +49,11 @@ const apiAdapter = require('../../common/api_adapter');
 const config = require('../../common/config');
 const mask = require('json-mask');
 
-const bearerAuth = 'bearer abc123';
+const terraAuth = 'bearer abc123';
 
 const mockRequest = (req, requestFields = MARTHA_V3_ALL_FIELDS) => {
     req.method = 'POST';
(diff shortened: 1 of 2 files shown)


test/martha/martha_v3.test.js
```diff
@@ -745,34 +761,28 @@ test.serial('martha_v3 parses a Gen3 CRDC CIB URI response correctly', async (t)
 
 // BT-236 temporarily cut access token and access endpoint out of the flow
 test.serial('martha_v3 parses BDC response correctly', async (t) => {
+    const drsHost = config.HOST_BIODATA_CATALYST_STAGING;
+    const {
+        id: objectId, self_uri: drsUri,
+        access_methods: { 0: { access_id: accessId, access_url: { url: gsUrl } } }
+    } = bdcDrsResponse;
```
kshakir, line 768: NOTE: This destructuring of metadata responses will not work for `self_uri` from `hcaDrsResponse` and some others. (Yay "specifications"!) But those tests can be modified to hard code the `drsUri` some other way.

kshakir (approved):
Much simpler! 🎉

mcovarr (approved):

PR #223, opened 2021-06-01
Test refactoring [BT-302]

breilly2:
Related to (but does not complete) [BT-259](https://broadworkbench.atlassian.net/browse/BT-259).

At a high level, the pattern you will see is that, instead of checking that all of the arguments passed to the `getJsonFrom` stub at the end of the test, the stub is configured to only respond when the arguments are what the test case is expecting. This both reduces the line count in most cases and allows us to be less specific about the order that we expect the calls to happen in (though we could still explicitly check call order in some cases if we need to).

I've added comments in this PR to describe a bit of my thought process. This was more work than I expected and was done over many separate sessions, so I probably didn't capture anything, but it should at least give you a snapshot of my thinking. Please read each of my PR comments, consider them, and bring up anything that you think of along the way.


test/martha/martha_v3.test.js
```diff
@@ -113,25 +136,20 @@ test.serial('martha_v3 uses the default error handler for unexpected errors', as
...
-    t.falsy(getJsonFromApiStub.getCall(0).args[1]); // no auth passed
-    const requestedBondSAKeyUrl = getJsonFromApiStub.getCall(1).args[0];
-    const saKeyUrlMatches = requestedBondSAKeyUrl.match(bondSAKeyUrlRegEx);
-    t.truthy(saKeyUrlMatches, 'Bond SA key URL called does not match Bond SA key URL regular expression');
-    const expectedSAKeyProvider = 'dcf-fence';
-    const actualSAKeyProvider = saKeyUrlMatches[2];
-    t.is(actualSAKeyProvider, expectedSAKeyProvider);
+    t.deepEqual(result, sampleDosMarthaResult(googleSAKeyObject));
+
+    sinon.assert.callCount(getJsonFromApiStub, 2);
```
breilly2, line 157: I would rather use sinon mocks (instead of stubs) and `verify()` them here, but I couldn't get mocks to work. Checking the call count, in combination with the defined stub expectations, gets most of what we want. I think it would just be easier to understand failures if we could use `verify()`.

test/martha/martha_v3.test.js
```diff
@@ -141,155 +159,121 @@ test.serial('martha_v3 resolves a valid DOS-style url', async (t) => {
...
-    t.truthy(saKeyUrlMatches, 'Bond SA key URL called does not match Bond SA key URL regular expression');
-    const expectedSAKeyProvider = 'dcf-fence';
-    const actualSAKeyProvider = saKeyUrlMatches[2];
-    t.is(actualSAKeyProvider, expectedSAKeyProvider);
 });
 
-test.serial("martha_v3 resolves successfully and ignores extra data submitted besides a 'url'", async (t) => {
-    getJsonFromApiStub.onCall(0).resolves(sampleDosResponse);
-    getJsonFromApiStub.onCall(1).resolves(googleSAKeyObject);
+test.serial("martha_v3 doesn't fail when extra data submitted besides a 'url'", async (t) => {
```
breilly2, line 181: This test had a lot of overlap with the test above. I'm wondering if we can still get value while being much less explicit about the expectations instead of duplicating the entire test.

test/martha/martha_v3.test.js
```diff
@@ -141,155 +159,121 @@ test.serial('martha_v3 resolves a valid DOS-style url', async (t) => {
...
+
+    getJsonFromApiStub.withArgs(bond.serviceAccountKeyUrl, terraAuth).resolves(googleSAKeyObject);
+    getJsonFromApiStub.withArgs(dos.dataobjectsUrl, null).resolves(sampleDosResponse);
+
     const response = mockResponse();
     await marthaV3(mockRequest({ body: { 'url': 'drs://abc/123' } }), response);
+
     t.is(response.statusCode, 200);
     const result = response.send.lastCall.args[0];
+    t.deepEqual(result, sampleDosMarthaResult(googleSAKeyObject));
```
breilly2: I can't remember if there was a specific reason why we might need to `{ ...result }` here. I've been changing these as I go and the tests continue to pass. Is there something I'm not thinking of that could cause this to give the wrong signal?

test/martha/martha_v3.test.js
```diff
@@ -141,155 +159,121 @@ test.serial('martha_v3 resolves a valid DOS-style url', async (t) => {
...
-    );
+    await marthaV3(mockRequest({ body: { url: drsUri, fields: ['accessUrl'] } }), response);
+
     t.is(response.statusCode, 200);
     const result = response.send.lastCall.args[0];
-    sinon.assert.callCount(getJsonFromApiStub, 1);
-    t.deepEqual(
-        { ...result },
-        { accessUrl: null });
+    t.deepEqual(result, { accessUrl: drsAccessUrlResponse });
```
breilly2: This is a change in the expectation. This was previously using a BDC CIB DRS URI, but I switched it to KidsFirst since BDC is not yet turned on.

test/martha/martha_v3.test.js
```diff
@@ -644,34 +628,28 @@ test.serial('martha_v3 parses a Gen3 CRDC CIB URI response correctly', async (t)
...
+    const drsHost = config.HOST_BIODATA_CATALYST_STAGING;
+    const {
+        id: objectId, self_uri: drsUri,
+        // access_methods: { 0: { access_id: accessId, access_url: { url: gsUrl } } }
+    } = bdcDrsResponse;
+    const bond = bondUrls('fence');
+    const drs = drsUrls(drsHost, objectId);
+    // const drsAccessUrlResponse = mockGcsAccessUrl(gsUrl);
+
     // https://lucid.app/lucidchart/428a0bdd-a884-4fc7-9a49-7bf300ef6777/edit?shared=true&page=0_0#
```
breilly2: Move this comment to docs for `marthav3()` and add context.

test/martha/martha_v3.test.js
```diff
@@ -63,7 +65,16 @@ const mockResponse = () => {
...
+        send: sinon.mock('send').once().callsFake(function setBody(body) {
+            // Express will effectively JSON.stringify objects passed to send, which is perfect for
+            // sending over the wire and deserializing into a simple object on the other end.
+            // We want a similar effect here, where all we get are the object properties (and
+            // specifically no class instance details) so that we can make comparisons against
+            // simple objects. Specifically, this takes care of cases where we "send" a
+            // FailureResponse.
+            this.body = { ...body };
+            return this;
+        }),
```
breilly2, line 77: This lets us check `response.body` instead of
```
const result = response.send.lastCall.args[0]
{ ...result }
```
in our tests.

test/martha/martha_v3.test.js
```diff
@@ -85,21 +120,16 @@ function mockS3AccessUrl(s3UrlString) {
     return { url: `https://${s3Url.hostname}.s3-website.us-west-2.amazonaws.com${s3Url.pathname}?sig=ABC` };
 }
 
-const bondSAKeyUrlRegEx = /^https:\/\/([^/]+)\/api\/link\/v1\/([a-z-]+)\/serviceaccount\/key$/;
-
-const bondAccessTokenUrlRegEx = /^https:\/\/([^/]+)\/api\/link\/v1\/([a-z-]+)\/accesstoken$/;
```
breilly2, line 90: The tests now expect requests to specific endpoints, so we no longer need to after-the-fact check the args against these expressions.

test/martha/martha_v3.test.js
```diff
@@ -141,255 +164,188 @@ test.serial('martha_v3 resolves a valid DOS-style url', async (t) => {
...
-    const actualSAKeyProvider = saKeyUrlMatches[2];
-    t.is(actualSAKeyProvider, expectedSAKeyProvider);
-});
-
-test.serial("martha_v3 resolves successfully and ignores extra data submitted besides a 'url'", async (t) => {
-    getJsonFromApiStub.onCall(0).resolves(sampleDosResponse);
-    getJsonFromApiStub.onCall(1).resolves(googleSAKeyObject);
+    t.deepEqual(response.body, sampleDosMarthaResult(googleSAKeyObject));
+
+    sinon.assert.callCount(getJsonFromApiStub, 2);
```
breilly2, line 178: You might notice that I removed the trailing line comments from these call count assertions. My reasoning is that the number of expected calls is expressed in the `withArgs` stubbing at the top of the test.

If you'd rather keep these comments, let me know and I'll put them back in some form.

test/martha/martha_v3.test.js
```diff
@@ -112,26 +142,19 @@ test.serial('martha_v3 uses the default error handler for unexpected errors', as
 });
 
 // Test the "default" case because we don't know who you are.
-test.serial('martha_v3 resolves a valid DOS-style url', async (t) => {
-    getJsonFromApiStub.onCall(0).resolves(sampleDosResponse);
-    getJsonFromApiStub.onCall(1).resolves(googleSAKeyObject);
+test.serial('martha_v3 resolves a valid DOS-style url using the "dcf-fence" provider', async (t) => {
+    const bond = bondUrls('dcf-fence');
```
breilly2, line 146: A lot of the tests near the top of this file ask Martha about `dos://abc/123` which will fall through to the default case in `determineDrsType` which uses `dcf-fence`.

If we remove that default case, we will have to update several tests to use something a little more real (but can still probably be a contrived example instead of something mimicking an actual DRS object).

test/martha/martha_v3.test.js
```diff
@@ -424,412 +382,332 @@ test.serial('martha_v3 returns an error when an invalid field is requested', asy
...
-        const requestedBondSAKeyUrl = getJsonFromApiStub.getCall(1).args[0];
-        const saKeyUrlMatches = requestedBondSAKeyUrl.match(bondSAKeyUrlRegEx);
-        t.truthy(saKeyUrlMatches, 'Bond SA key URL called does not match Bond SA key URL regular expression');
-        const expectedSAKeyProvider = 'dcf-fence';
-        const actualSAKeyProvider = saKeyUrlMatches[2];
-        t.is(actualSAKeyProvider, expectedSAKeyProvider);
-    }
-);
+
+    t.is(response.statusCode, 200); // Not strictly necessary here, but the test fails if we don't assert something
```
breilly2, line 509: This also makes me feel like this test overlaps with other tests. I'm leaving it here because I'm envisioning a future where this case could be checked in a unit test suite (no mocks/stubs) and its presence will remind us that this needs to be covered in that test suite.

test/martha/martha_v3.test.js
```diff
@@ -424,412 +382,332 @@ test.serial('martha_v3 returns an error when an invalid field is requested', asy
...
+    sinon.assert.callCount(getJsonFromApiStub, 2);
 });
 
 test.serial('martha_v3 parses a Gen3 CRDC CIB URI response correctly', async (t) => {
-    getJsonFromApiStub.onCall(0).resolves(gen3CrdcResponse);
-    getJsonFromApiStub.onCall(1).resolves(googleSAKeyObject);
+    const bond = bondUrls('dcf-fence');
+    // TODO: This object ID is inconsistent with gen3CrdcResponse but, for now, it doesn't break
+    // anything. Eventually, when we turn on signed URLs for CRDC, we may want to reconcile the
+    // difference so that we can avoid duplication of these IDs between the tests and test data.
```
breilly2, line 568: Note this. I resolved some inconsistencies with BDC tests and test data, but not for other projects. I also didn't comprehensively audit and flag all of the inconsistencies in comments like this, so there are probably more to uncover.

TOL: I like that we have some set of tests that operates against real data that we've captured from real DRS servers. I would love to move the majority of the coverage we have today into more targeted unit tests that can use contrived test data that is more expressive of exactly what Martha behavior is being tested. Those tests should be a lot more self-contained, easier to read, and less prone to accidental breakage.

test/martha/martha_v3.test.js
```diff
@@ -424,412 +382,332 @@ test.serial('martha_v3 returns an error when an invalid field is requested', asy
...
-    const actualSAKeyProvider = saKeyUrlMatches[2];
-    t.is(actualSAKeyProvider, expectedSAKeyProvider);
+    sinon.assert.callCount(getJsonFromApiStub, 2);
 });
 
 test.serial('martha_v3 parses a The AnVIL CIB URI response correctly', async (t) => {
-    getJsonFromApiStub.onCall(0).resolves(anvilDrsResponse);
-    getJsonFromApiStub.onCall(1).resolves(googleSAKeyObject);
+    const bond = bondUrls('anvil');
+    const drs = drsUrls(config.HOST_THE_ANVIL_STAGING, 'dg.ANV0%2F00008531-03d7-418c-b3d3-b7b22b5381a0');
```
breilly2: This had me stuck for a bit until I realized I could use `%2F` here instead of `/`. It feels a little arbitrary to me, but I think that's because I don't know all of the history. There's a lot of special case code in the compact identifier resolution, and I would love to write some specific unit tests against that (which, again, would eliminate the need to some of the extra test permutations in this file).

test/martha/martha_v3.test.js
```diff
@@ -424,412 +382,332 @@ test.serial('martha_v3 returns an error when an invalid field is requested', asy
...
-    getJsonFromApiStub.onCall(0).resolves(kidsFirstDrsResponse);
-    getJsonFromApiStub.onCall(1).resolves(bondAccessTokenResponse);
-    getJsonFromApiStub.onCall(2).resolves(drsAccessUrlResponse);
+    const {
+        id: objectId, self_uri: drsUri,
+        access_methods: { 0: { access_id: accessId, access_url: { url: s3Url } } },
+    } = kidsFirstDrsResponse;
+    const drsAccessUrlResponse = mockS3AccessUrl(s3Url);
+    const bond = bondUrls('kids-first');
+    const drs = drsUrls(config.HOST_KIDS_FIRST_PROD, objectId, accessId);
```
breilly2: I think I did this in other places as well, but this is a case where I had to change the host from `STAGING` to `PROD`... I think in order to match the hosts encoded in the test data. I _think_ this is okay, but please take a minute to consider this while reviewing and reply here with your thoughts. Thanks.

test/martha/martha_v3.test.js
```diff
@@ -424,412 +382,332 @@ test.serial('martha_v3 returns an error when an invalid field is requested', asy
...
-    const drsAccessUrlResponse = mockS3AccessUrl(kidsFirstDrsResponse.access_methods[0].access_url.url);
-    getJsonFromApiStub.onCall(0).resolves(kidsFirstDrsResponse);
-    getJsonFromApiStub.onCall(1).resolves(bondAccessTokenResponse);
-    getJsonFromApiStub.onCall(2).resolves(drsAccessUrlResponse);
+    const {
+        id: objectId,
+        access_methods: { 0: { access_id: accessId, access_url: { url: s3Url } } }
+    } = kidsFirstDrsResponse;
+    const bond = bondUrls('kids-first');
+    const drs = drsUrls(config.HOST_KIDS_FIRST_STAGING, objectId, accessId);
```
breilly2, line 690: I didn't have to change this one from `STAGING` to `PROD` for some reason? 🤷

test/martha/martha_v3.test.js
kshakir: ToL: If anything I've heard rumors that the default case will _expand_. We can ask Kyle for more info.

The theory being that "DRS is a fully realized spec and any and all DRS providers should behave the same way".

test/martha/martha_v3.test.js
kshakir: ToL: If one is ever looking to try again, Leaving a [link here](https://github.com/broadinstitute/martha/blob/2cd644f8e9ee026174f38cd0af12145a2abfbe17/common/createSignedGsUrl.js#L3-L7) to a previous battle with sinon.

test/martha/martha_v3.test.js
kshakir: ToL: Due to a lack of \*real\* integration tests extra care was put into unit test expectations. I believe this is one regression test case ensuring that we don't break clients if they add extra data. NOTE: We do later return errors if one requests a `field` that cannot be returned.

One day we will use the glory of RAS to test DRS providers beyond GCS/MockDRS.

test/martha/martha_v3.test.js
```diff
@@ -141,255 +164,188 @@ test.serial('martha_v3 resolves a valid DOS-style url', async (t) => {
...
+
+    sinon.assert.callCount(getJsonFromApiStub, 1);
 });
 
-// BT-236 temporarily cut access token and access endpoint out of the flow
 test.serial('martha_v3 calls the correct endpoints when only the accessUrl is requested', async (t) => {
-    getJsonFromApiStub.onCall(0).resolves(bdcDrsResponse);
+    const {
+        id: objectId, self_uri: drsUri,
+        access_methods: { 0: { access_id: accessId, access_url: { url: s3Url } } }
```
kshakir, line 230: TiL: Wow- never thought of destructuring an array using `{ 0: ... }` 👍
ToL: Would something like this work too, or is this not valid ecmascript?

```javascript
access_methods: [ { access_id: accessId, access_url: { url: s3Url } } ]
```

test/martha/martha_v3.test.js
kshakir: The `%2F` vs. `/` is a real example. 😞

I can re-dig it up, but they are different inputs given to us in some google doc somewhere. The list of docs is elsewhere in the martha codebase.

test/martha/martha_v3.test.js
kshakir: If this was like most non-TDR DRS hosts, their prod servers probably existed before test/staging servers. So I suspect that the test data was generated from a prod example. Now that staging exists maybe a new example can be generated for the unit tests.

test/martha/martha_v3.test.js
kshakir: The hardcoded CIB URI with `dg.F82A1A` is expanding to the `STAGING` host. The previous test was using host-based URIs extracted from one of the hardcoded responses and feeding that URI \*into\* the test's martha_v3 request.

kshakir (approved):

test/martha/martha_v3.test.js
breilly2: Oh right, I found it: `FailureResponse`. See changes to `mockRequest` in this PR.

test/martha/martha_v3.test.js
breilly2: Right, though the behavior of the default case will change. Right now, we're kinda emulating martha_v2, but in the future, it'll be a more generalized martha_v3 (with no bond provider, I think, because we wouldn't know which one to use).

test/martha/martha_v3.test.js
breilly2: Ah, good point! Sounds like a good idea to me, now that (I'm pretty sure) there is a Kids First staging server.

test/martha/martha_v3.test.js
breilly2: I think I tried something like that before looking it up and finding the `{ 0: ... }` technique.

mcovarr (approved):

PR #224, opened 2021-06-16
Don't fail from martha_v3 if there's a problem fetching a signed URL [BT-259]

breilly2:
After the previous test refactoring, this completes https://broadworkbench.atlassian.net/browse/BT-259.

This code is... kinda ugly. I don't really like it, but it'll get the job done. Eventually, errors fetching signed URLs will be fatal and most/all of the ugliness added in this PR will go away.

### DIFF
--- a/martha/martha_v3.js
+++ b/martha/martha_v3.js
@@ -593,21 +593,21 @@ async function retrieveFromServers(params) {
         }
     };
 
-    let ranOutOfTime;
+    const timeout = Symbol.for('timeout');
     const waitUntilTimeIsAlmostUp = async () => {
         await delay(pencilsDownSeconds * 1000);
-        ranOutOfTime = true;
+        return timeout;
     };
 
     // For now, since we can still return native object URLs, we don't want to fail if it takes too
     // long to fetch a signed URL. Eventually, we won't have native object URLs to fall back on so
     // we'll have to throw an error, at which point we can remove all of this `Promise.race` stuff.
-    await Promise.race([
+    const winner = await Promise.race([
         fetch(),
         waitUntilTimeIsAlmostUp(),
     ]);
 
-    if (ranOutOfTime && !hypotheticalErrorMessage) {
+    if (winner === timeout && !hypotheticalErrorMessage) {
         console.log('Ran out of time fetching a signed URL; returning the native URL instead of throwing an error.');
     }
 
--- a/martha/martha_v3.js
+++ b/martha/martha_v3.js
@@ -5,6 +5,7 @@ const {
     RemoteServerError,
     logAndSendBadRequest,
     logAndSendServerError,
+    delay,
 } = require('../common/helpers');
 const config = require('../common/config');
 const apiAdapter = require('../common/api_adapter');
@@ -86,6 +87,16 @@ const DG_COMPACT_BDC_STAGING = 'dg.712c';
 const DG_COMPACT_THE_ANVIL = 'dg.anv0';
 const DG_COMPACT_CRDC = 'dg.4dfc';
 const DG_COMPACT_KIDS_FIRST = 'dg.f82a1a';
+
+// Google Cloud Functions gives us 60 seconds to respond. We'll use most of it to fetch what we can,
+// but not fail if we happen to time out when fetching a signed URL takes too long.
+let pencilsDownTime = 58;
+
+// For tests that need to probe the behavior near the limit of the Cloud Function timeout, we don't
+// want to have to wait 60 seconds.
+const overridePencilsDownTime = (seconds) => {
+    pencilsDownTime = seconds;
+};
 
 // noinspection JSUnusedGlobalSymbols
 class DrsType {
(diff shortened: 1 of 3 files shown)


martha/martha_v3.js
```diff
@@ -87,6 +88,16 @@ const DG_COMPACT_THE_ANVIL = 'dg.anv0';
 const DG_COMPACT_CRDC = 'dg.4dfc';
 const DG_COMPACT_KIDS_FIRST = 'dg.f82a1a';
 
+// Google Cloud Functions gives us 60 seconds to respond. We'll use most of it to fetch what we can,
+// but not fail if we happen to time out when fetching a signed URL takes too long.
+let pencilsDownTime = 58;
```
breilly2: I'm not in love with this variable name, but I was struggling to think of something more conventional. This metaphor made sense to me. Let me know if you don't think this is clear or if you have other suggestions.

martha/martha_v3.js
```diff
@@ -87,6 +88,16 @@ const DG_COMPACT_THE_ANVIL = 'dg.anv0';
...
+
+// For tests that need to probe the behavior near the limit of the Cloud Function timeout, we don't
+// want to have to wait 60 seconds.
+const overridePencilsDownTime = (seconds) => {
```
breilly2: 🤢 , but it's better than making each test that needs to bump against the timeout (currently 2 test cases) take a minute to run.

martha/martha_v3.js
```diff
@@ -478,87 +489,130 @@ async function retrieveFromServers(params) {
...
-    if (overlapFields(requestedFields, MARTHA_V3_METADATA_FIELDS)) {
+    let bondSA;
+    let drsResponse;
+    let fileName;
+    let accessUrl;
+
+    // `fetch` below might time out in a way that we want to report as an error. Since the timeout
+    // interrupts us while we're waiting for a response, we need to capture what we were about to
+    // try doing just before we do it so that we can provide that detail in the error report.
+    let hypotheticalErrorMessage;
```
breilly2, line 500: 🤢 🤢 , but I couldn't think of a better way to surface this detail. Again, this will go away once failure to fetch a signed URL is fatal.

martha/martha_v3.js
```diff
@@ -478,87 +489,130 @@ async function retrieveFromServers(params) {
...
+
+    const fetch = async () => {
```
breilly2, line 502: Best viewed ignoring whitespace. The changes here are:
1. Wrapping all of the server fetching into a `fetch` function
2. Moving the SA key fetch from the bottom to the top
3. Pulling all of the `let`s out of the function so they can be accessed after calling the function
4. Managing `hypotheticalErrorMessage` (🤢 )
5. Wrapping everything involved in fetching a signed URL in a `try`/`catch` to log instead of throwing

martha/martha_v3.js
```diff
@@ -478,87 +489,130 @@ async function retrieveFromServers(params) {
...
-    if (bondProvider && accessId) {
-        try {
-            const httpsAccessUrl = generateAccessUrl(drsType, accessId);
-            const accessTokenAuth = `Bearer ${accessToken}`;
-            console.log(`Requesting DRS access URL for '${url}' from '${httpsAccessUrl}'`);
-            accessUrl = await apiAdapter.getJsonFrom(httpsAccessUrl, accessTokenAuth);
-        } catch (error) {
-            throw new RemoteServerError(error, 'Received error contacting DRS provider.');
-        }
+    let ranOutOfTime;
```
breilly2: I was getting used to `let`s last night and forgot that I could instead return something from this function (and `fetch`) to indicate which promise won the race. I'll probably add a commit to this PR to clean this up.

test/martha/martha_v3.test.js
```diff
@@ -458,8 +460,6 @@ test.serial('martha_v3 should return 500 if Data Object resolution fails', async
     t.is(response.statusCode, 500);
     t.is(response.body.status, 500);
     t.is(response.body.response.text, 'Received error while resolving DRS URL. Data Object Resolution forced to fail by testing stub');
-
-    sinon.assert.callCount(getJsonFromApiStub, 1);
```
breilly2, line 462: For this case (and the 2 below), I'm proposing that we don't really care how many times the stub was called as long as the error that the stub throws is represented in the response. Please speak up if you disagree. This was fallout from changing the order of fetching the SA key.

test/martha/martha_v3.test.js
```diff
@@ -728,6 +724,83 @@ test.serial('martha_v3 parses HCA response correctly', async (t) => {
...
+test.serial('martha_v3 succeeds if an error is encountered while fetching a signed URL', async (t) => {
+    const {
+        id: objectId, self_uri: drsUri,
+        access_methods: { 0: { access_id: accessId } }
+    } = kidsFirstDrsResponse;
+    const bond = bondUrls('kids-first');
+    const drs = drsUrls(config.HOST_KIDS_FIRST_STAGING, objectId, accessId);
+    getJsonFromApiStub.withArgs(drs.objectsUrl, null).resolves(kidsFirstDrsResponse);
+    getJsonFromApiStub.withArgs(bond.accessTokenUrl, terraAuth).resolves(bondAccessTokenResponse);
+    getJsonFromApiStub.withArgs(drs.accessUrl, `Bearer ${bondAccessTokenResponse.token}`)
```
breilly2, line 796: I discovered that this test is finicky in that it will still pass even if this stub doesn't get invoked. (I discovered this because I accidentally put a `:` after `Bearer` at first and the test didn't fail when I expected it to. Yay TDD!) I'm going to think about it a little more, but I think I'm going to have to add some assertions that the stub was actually invoked... much like the checks I removed during the test refactoring.

martha/martha_v3.js
```diff
@@ -478,87 +489,130 @@ async function retrieveFromServers(params) {
...
+    const waitUntilTimeIsAlmostUp = async () => {
+        await delay(pencilsDownTime * 1000);
```
mcovarr: units might be helpful in the variable name, i.e. `pencilsDownSeconds`

martha/martha_v3.js
```diff
@@ -478,87 +489,130 @@ async function retrieveFromServers(params) {
...
+    // For now, since we can still return native object URLs, we don't want to fail if it takes too
+    // long to fetch a signed URL. Eventually, we won't have native object URLs to fall back on so
+    // we'll have to throw an error, at which point we can remove all of this `Promise.race` stuff.
+    await Promise.race([
+        fetch(),
+        waitUntilTimeIsAlmostUp(),
+    ]);
+
+    if (ranOutOfTime && !hypotheticalErrorMessage) {
+        console.log('Ran out of time fetching a signed URL, but returning everything else we fetched instead of throwing an error.');
```
mcovarr: TOL The wording here might lead to some puzzlement... maybe"returning native URL access data" or something like that?

test/martha/martha_v3.test.js
mcovarr: IMHO this seems fine for these unhappy path cases since the error message is checked specifically

mcovarr (approved):

test/martha/martha_v3.test.js
kshakir: I don't have strong opinions on this test change right now so LGTM. 👍

And if I change my mind I can always (file a ticket, refine, backlog, select for sprint, and then...) submit a PR.

test/martha/martha_v3.test.js
```diff
@@ -728,6 +724,83 @@ test.serial('martha_v3 parses HCA response correctly', async (t) => {
...
+const testWithTimeout = (ms, asyncTestFn) => (t) => {
+    const waitThenFail = async () => {
+        await delay(ms);
+        t.fail(`Test did not complete within ${ms}ms`);
+    };
+    return Promise.race([asyncTestFn(t), waitThenFail()]);
+};
+
+test.serial('martha_v3 succeeds within 60 seconds even if fetching a signed URL never returns', testWithTimeout(5 * 1000, async (t) => {
+    overridePencilsDownTime(3);
```
kshakir: ToL: Is the "60 seconds" still applicable with the override? Perhaps drop it, leaving "succeeds even if fetching..."

kshakir (approved):
So much complexity.

But if ~DOS~ DRS were simple, this entire git repo wouldn't exist. 👍

martha/martha_v3.js
breilly2: I went with "everything else" because the request might not have asked for any fields from DRS, like only requesting an SA key... but now that I type that, I realize that we wouldn't hit this case unless we were trying to get a signed URL which implies that we fetched the rest of the DRS metadata, so 👍 .

test/martha/martha_v3.test.js
breilly2: Oh yeah, I noticed that at one point and then forgot to get back to it. I'll update the test name and probably add a clarifying comment.

PR #225, opened 2021-07-08
Fail when asked to resolve dataguids.org DRS URIs [BT-320]

breilly2:
https://broadworkbench.atlassian.net/browse/BT-320

Since this is an emergency fix, instead of removing the default case, I chose to specifically identify the dataguids.org case and return a (hopefully) helpful error.

### DIFF
--- a/martha/martha_v3.js
+++ b/martha/martha_v3.js
@@ -407,6 +407,11 @@ function determineDrsType(url) {
         );
     }
 
+    // RIP dataguids.org
+    if (host.endsWith('dataguids.org')) {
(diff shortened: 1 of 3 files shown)


mcovarr (approved):
LGTM

salonishah11 (approved):
LGTM 👍  
Should there be a JIRA ticket associated to this PR?

breilly2:
@salonishah11 Yes, there should! Thank you for catching that. I updated the title and description.

Also worth noting that the urgency is because it seems to possibly be causing another occurrence of https://broadworkbench.atlassian.net/browse/BT-313.

PR #227, opened 2021-07-14
Unknown provider is error [BT-199]

mcovarr:
~Just a draft for now, unit tests are passing but coverage is lacking and integration tests are red.~ Review away!

### DIFF
--- a/test/martha/martha_v3.test.js
+++ b/test/martha/martha_v3.test.js
@@ -175,13 +175,13 @@ test.serial("martha_v3 doesn't fail when extra data submitted besides a 'url'", 
 });
 
 test.serial('martha_v3 does not call Bond when only DRS fields are requested', async (t) => {
-    const dos = dosUrls('abc', '123');
-    getJsonFromApiStub.withArgs(dos.dataobjectsUrl, null).resolves(sampleDosResponse);
+    const drs = drsUrls(bdc, '123');
+    getJsonFromApiStub.withArgs(drs.objectsUrl, null).resolves(sampleDosResponse);
     const response = mockResponse();
 
     await marthaV3(mockRequest({
         body: {
-            url: 'dos://abc/123',
+            url: `dos://${bdc}/123`,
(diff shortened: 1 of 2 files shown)


kshakir (commented):
Repeating what the bot said: some of the new errors do not seem to have unit tests.

kshakir (approved):

test/martha/martha_v3.test.js
```diff
@@ -965,24 +946,25 @@ function determineDrsTypeTestWrapper(testUrl) {
...
 test.serial('martha_v3 determineDrsType should parse drs:// Data Object uri with query part', (t) => {
     t.is(
-        determineDrsTypeTestWrapper('drs://fo.o/bar?version=1&bananas=yummy'),
-        'https://fo.o/ga4gh/dos/v1/dataobjects/bar?version=1&bananas=yummy'
+        determineDrsTypeTestWrapper(`drs://${bdc}/bar?version=1&bananas=yummy`),
+        `https://${bdc}/ga4gh/drs/v1/objects/bar?version=1&bananas=yummy`
     );
 });
 
 test.serial('martha_v3 determineDrsType should parse drs:// Data Object uri when host includes a port number', (t) => {
```
breilly2, line 963: Looks like this is actually illegal, according to the spec (https://ga4gh.github.io/data-repository-service-schemas/preview/release/drs-1.1.0/docs/#_compact_identifier_based_drs_uris):
> The colon character is not allowed in a Hostname-based DRS URI, making it easy to tell them apart.

breilly2 (approved):
👍  Looks good! Glad to finally be doing this!

test/martha/martha_v3.test.js
mcovarr: Uhh not sure what to do here as I suspect this test was added to accommodate a real world example... the test treats the number after the colon as a port but the docs are saying that's actually an 'accession'.  🙁

mcovarr: Written up as BT-345.

PR #232, opened 2021-09-09
Signed URL build failures should be errors for S3 objects [BT-392]

mcovarr:


mcovarr:
jenkins retest

test/martha/martha_v3.test.js
```diff
@@ -815,7 +815,7 @@ test.serial('martha_v3 fails if something times out before trying to fetch a sig
     );
 }));
 
-test.serial('martha_v3 succeeds if an error is encountered while fetching a signed URL', async (t) => {
```
breilly2, line 818: I think we still want this test, though I agree that we can add another test for the S3 case.

breilly2 (changes requested):
Requesting that we add a new test instead of replacing an existing test. Otherwise, looks good.

test/martha/martha_v3.test.js
mcovarr: In its current state the Martha v3 code does not attempt to fetch signed URLs for non-S3 objects, so there can't be errors in that process.

breilly2: Ugh. I guess you're right.

I'm frustrated because I put a bunch of work into adding the behavior that this test checks. As the test is now, I'm afraid all of that work could be undone and all the tests will still pass.

One of these days, I'd like to rewrite these tests to check for the expected behavior based on different permutations of `DrsType` instead of specifically for the projects we support. (We probably want some tests like that too, but we could be testing the base capabilities of Martha separately.)

breilly2 (approved):

martha/martha_v3.js
```diff
@@ -626,7 +626,13 @@ async function retrieveFromServers(params) {
                 }
             }
         } catch (error) {
-            // Just logging the error for now. Eventually, we'll want to remove this outer try.
+            if (accessMethodTypeIsS3) {
+                // Throw if this is S3 and we failed to get a signed URL. Martha has no concept of a native S3
+                // path so the caller will not have a fallback means of accessing the object.
+                throw error;
```
aednichols, line 632: TOL

It can be hard to track down exactly where errors were thrown, and whether they came from library code or ours. If we have to throw, maybe instantiate a new error with some custom context like we do on line 653 (659)  below?

aednichols (approved):

martha/martha_v3.js
mcovarr: This error will have custom context since it is a re-throw of one of the wrapper `RemoteServerError`s generated on line 600, 613 or 625. The [martha v3 handler](https://github.com/broadinstitute/martha/blob/aaecfb5cd9f41d383823a88133d58ae9086a5c84/martha/martha_v3.js#L722) knows about these wrapper`RemoteServerError`s and will extract the `cause` as a parameter to `logAndSendServerError`. If that code finds a `status` [property on the underlying error](https://github.com/broadinstitute/martha/blob/aaecfb5cd9f41d383823a88133d58ae9086a5c84/common/helpers.js#L468) its value will be assigned into the HTTP response, which of course is important information for callers of Martha to have in deciding whether to retry their requests. `logAndSendServerError` also extracts as much useful `description` and `message` context as it can to return to the caller.

PR #235, opened 2021-10-05
Refactor to allow for getting signed URLs from TDR [BT-406]

mcovarr:
Allows for passing the `martha-force-access-url` header to force Martha to retrieve a signed URL even if the DRS provider is not configured to ordinarily return signed URLs. Also a fair bit of other, hopefully simplifying refactoring.

### DIFF
--- a/martha/drs_providers.js
+++ b/martha/drs_providers.js
@@ -8,7 +8,7 @@ const config = require("../common/config");
 const {
     MARTHA_V3_ACCESS_ID_FIELDS,
     MARTHA_V3_BOND_SA_FIELDS,
-    overlapFields
+    overlapFields, MARTHA_V3_METADATA_FIELDS
 } = require("./martha_fields");
 
 const AccessMethodType = {
@@ -54,7 +54,7 @@ class DrsProvider {
         this.bondProvider = bondProvider;
         this.accessMethods = accessMethods;
         this.accessMethodTypes = accessMethods && accessMethods.map((m) => m.accessMethodType);
-        // may be overridden in request headers or tests
+        // may be overridden in request headers or tests, set it explicitly here to placate eslint.
         this.forceAccessUrl = false;
     }
 
@@ -70,7 +70,7 @@ class DrsProvider {
      * @param requestedFields
      * @return {boolean}
      */
-    shouldFetchFenceToken(accessMethod, requestedFields) {
+    shouldFetchFenceAccessToken(accessMethod, requestedFields) {
         return this.bondProvider &&
             overlapFields(requestedFields, MARTHA_V3_ACCESS_ID_FIELDS) &&
             (this.forceAccessUrl || (accessMethod &&
@@ -109,10 +109,21 @@ class DrsProvider {
             overlapFields(requestedFields, MARTHA_V3_BOND_SA_FIELDS);
     }
 
-    shouldFailOnAccessUrlFail(accessMethod) {
-        // Fail if we failed to get a signed URL and the access method is truthy but its type is not GCS. Martha clients
-        // currently can't deal with cloud paths other than GCS so there isn't a fallback way of accessing the object.
-        return this && accessMethod && accessMethod.type !== AccessMethodType.GCS;
+    static shouldFailOnAccessUrlFail(accessMethod) {
+        // Fail this request if Martha was unable to get an access/signed URL and the access method is truthy but its
+        // type is not GCS. Martha clients currently can't deal with cloud paths other than GCS so there isn't a
+        // fallback way of accessing the object.
+        // Note: TDR's metadata responses for objects stored in GCS include an `https` access method with a URL and
+        // headers (the headers containing the same bearer auth as in the TDR metadata request) which could be used for
+        // download without fetching a signed URL. However this presumes that the URL downloaders in Martha clients
+        // support headers, which at the time of this writing is not true at least for the Cromwell localizer using getm
+        // 0.0.4. This also presumes that the Martha response would fall back to a different access method than the
+        // GCS/Azure one for which Martha tried and failed to get a signed URL. The current code does not support this.
+        return accessMethod && accessMethod.type !== AccessMethodType.GCS;
+    }
+
+    static shouldRequestMetadata(requestedFields) {
+        return overlapFields(requestedFields, MARTHA_V3_METADATA_FIELDS);
     }
 
     accessUrlAuth(accessMethod, accessToken, requestAuth) {
@@ -130,13 +141,13 @@ class DrsProvider {
 }
 
 // Define the default provider implementations here and insert them as the values in the `DrsProviderInstances` object
-// below. Test should override the map values to ensure providers work as expected with signed URLs turned on or off.
+// below. Tests can override the map values to ensure providers work as expected with signed URLs on or off.
 const BioDataCatalystDrsProvider = new DrsProvider(
     'BioData Catalyst (BDC)',
     MetadataAuth.NO,
     BondProvider.FENCE,
     [
-        //  BT-236 BDC signed URLs temporarily turned off
+        //  BT-236 BDC access (signed) URLs temporarily turned off
         new AccessMethod(AccessMethodType.GCS, AccessUrlAuth.FENCE_TOKEN, FetchAccessUrl.NO)
     ]
 );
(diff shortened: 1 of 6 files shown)


mcovarr:
jenkins retest

martha/martha_v3.js
```diff
@@ -471,30 +289,23 @@ function validateRequest(url, auth, requestedFields) {
...
         auth,
+        forceAccessUrl,
     } = params;
 
     validateRequest(url, auth, requestedFields);
-    const drsType = determineDrsType(url);
+    const urlParts = getHttpsUrlParts(url);
+    const drsProvider = determineDrsProvider(url, urlParts);
+    // Force the retrieval of a (signed) access URL for this request if the `martha-force-access-url` header is set.
+    Object.assign(drsProvider, {forceAccessUrl: Boolean(forceAccessUrl)});
```
breilly2: Since `determineDrsProvider` reuses object instances, and since I'm pretty sure Cloud Function instances can service multiple requests, using `Object.assign` here could leak `forceAccessUrl` from one request to another.

Oh, but maybe a Cloud Function instance only handles one request at a time, so as long as `Boolean(forceAccessUrl)` evaluates to something then`forceAccessUrl` will have the correct value for _this_ invocation. It still makes me a little uncomfortable. I'd rather we deal in disposable instances of `DrsProvider`.

Also, I prefer `!!forceAccessUrl` (which I think of as "coerce to boolean") instead of `Boolean(forceAccessUrl)`. The double-negative felt like a little bit of a dirty hack to me at first, but I got used to it pretty quickly. `Boolean(...)`, which is correct, looks too much like `new Boolean(...)`, which is incorrect; from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean:
```
var x = Boolean(expression);     // use this...
var x = !!(expression);          // ...or this
var x = new Boolean(expression); // don't use this!
```

test/martha/martha_v3.test.js
```diff
@@ -30,31 +30,41 @@ const {
...
 const apiAdapter = require('../../common/api_adapter');
 const config = require('../../common/config');
 const { delay } = require('../../common/helpers');
 const mask = require('json-mask');
+const {BondProviders} = require("../../common/bond");
 
 const terraAuth = 'bearer abc123';
 
-const mockRequest = (req, requestFields = MARTHA_V3_ALL_FIELDS) => {
+const mockRequest = (req, requestFields = MARTHA_V3_ALL_FIELDS, forceAccessUrl = false) => {
```
breilly2: Another way to do this would be to put the optional parameters an options object as the last parameter:
```
const mockRequest = (req, { fields = MARTHA_V3_ALL_FIELDS, forceAccessUrl = false }) => {
```
Then the caller would look like:
```
    const request = mockRequest(
        { body: { url: drsUri } },
        fields: MARTHA_V3_ACCESS_ID_FIELDS,
        forceAccessUrl: true
    );
```
To be fair,`fields` gets slightly more verbose, but I think `forceAccessUrl: true` is much better than having to have a `const FORCE_ACCESS_URL = true`.

martha/martha_v3.js
mcovarr: Yes it is my understanding that Cloud Functions only handles one request at a time. If the framework is actually reentrant we may need to audit our casual use of global variables...

`eslint` is configured in this repo with the `no-implicit-coercion` setting as described [here](https://eslint.org/docs/rules/no-implicit-coercion). I did write this with `!!` originally and was admonished to change it to `Boolean`. 🙂

test/martha/martha_v3.test.js
mcovarr: I'll have another look at cleaning this up. I don't love this syntax either but I really didn't want to change the other 40-odd invocations of `mockRequest` to clean up this one.

mcovarr:
The integration test failure here appears to be real, I'm looking at this as soon as I get my new machine all set up...

martha/martha_v3.js
breilly2: Boo. Okay.

What would you think of not reusing instances of `DrsProvider` to avoid any appearance of possibly bleeding state from one request to the next?

mcovarr: Yeah I'll look at not reusing these instances.

package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "start": "functions-framework --target=index --port=8010",
     "test": "./minnie-kenny.sh && nyc --reporter=html --reporter=text ava -m !smoketest* -m !live_test* -m !integration*",
+    "me": "ava --match 'martha_v3 calls the correct endpoints when access url fetch is forced for The AnVIL'",
```
jgainerdewar: This looks like it might be an accidental check-in.

jgainerdewar (approved):
This looks reasonable to me. As someone new to this codebase I like this refactor a lot, I think it helps clarify what Martha is doing.

mcovarr:
jenkins retest

martha/martha_v3.js
```diff
@@ -471,30 +290,27 @@ function validateRequest(url, auth, requestedFields) {
...
     validateRequest(url, auth, requestedFields);
-    const drsType = determineDrsType(url);
+    const urlParts = getHttpsUrlParts(url);
+
+    // Force the retrieval of a (signed) access URL for this request if the `martha-force-access-url` header is set
+    // copying the original object with `forceAccessUrl` set appropriately.
+    // Shallow clone a JavaScript ES6 class instance
+    // https://stackoverflow.com/questions/41474986/how-to-clone-a-javascript-es6-class-instance
+    const drsProvider = { ...determineDrsProvider(url, urlParts), forceAccessUrl: Boolean(forceAccessUrl) };
+    Object.setPrototypeOf(drsProvider, DrsProvider.prototype);
```
breilly2: I'm curious why not just have `determineDrsProvider` return a new instance every time, but this should also work.

martha/drs_providers.js
```diff
@@ -0,0 +1,251 @@
...
+    shouldFailOnAccessUrlFail(accessMethod) {
+        // Fail this request if Martha was unable to get an access/signed URL and the access method is truthy but its
+        // type is not GCS. Martha clients currently can't deal with cloud paths other than GCS so there isn't a
+        // fallback way of accessing the object.
+        // Note: TDR's metadata responses for objects stored in GCS include an `https` access method with a URL and
+        // headers (the headers containing the same bearer auth as in the TDR metadata request) which could be used for
+        // download without fetching a signed URL. However this presumes that the URL downloaders in Martha clients
+        // support headers, which at the time of this writing is not true at least for the Cromwell localizer using getm
+        // 0.0.4. This also presumes that the Martha response would fall back to a different access method than the
+        // GCS/Azure one for which Martha tried and failed to get a signed URL. The current code does not support this.
```
breilly2, line 123: 👍 ⭐

test/martha/_martha_v3_resources.js
```diff
@@ -155,6 +157,55 @@ const jadeDrsMarthaResult = {
     }
 };
 
+const jadeAccessUrlMetadataResponse = {
+    "id": "v1_0c86170e-312d-4b39-a0a4-2a2bfaa24c7a_c0e40912-8b14-43f6-9a2f-b278144d0060",
+    "name": "hapmap_3.3.hg38.vcf.gz",
+    // 2021-10-04 this is a minor forgery, dev TDR actually had a `null` value for the "self_uri" key
```
breilly2, line 163: 🤔

test/martha/martha_v3.test.js
```diff
@@ -976,36 +1084,18 @@ test.skip('martha_v3 should return 500 on exception trying to get signed URL fro
...
-    const result = getDrsAccessId(drsResponse, 'mismatching_access_method_type');
-    t.is(typeof result, 'undefined');
-});
-
-test.serial('martha_v3 getDrsAccessId should not return an access_id for a null accessMethodType', (t) => {
-    const drsResponse = { access_methods: [{ type: 'some_access_method_type', access_id: 'some_access_id' }] };
-    const result = getDrsAccessId(drsResponse, null);
-    t.is(typeof result, 'undefined');
-});
-
```
breilly2, line 996: Since the `getDrsAccessId` function went away, can I assume that these behaviors are already covered by other tests?

breilly2 (approved):
👍 Looks good overall. Nice refactoring! I'll admit that I've mostly not used classes in Javascript, so there are some places where I would have chosen different patterns.

test/martha/_martha_v3_resources.js
mcovarr: https://broadworkbench.atlassian.net/browse/DR-2156

PR #238, opened 2021-10-26
Add localizationPath [BT-417]

breilly2:
https://broadworkbench.atlassian.net/browse/BT-417

Add interpretation of `aliases` from TDR as a preferred localization path. Martha will return this in a new `localizationPath` field. `localizationPath` is included in the fields returned by default, which means that Martha responses from non-TDR DRS providers will always return `null` values for this.

### DIFF
--- a/martha/drs_providers.js
+++ b/martha/drs_providers.js
@@ -48,14 +48,14 @@ class AccessMethod {
 }
 
 class DrsProvider {
-    constructor(providerName, sendMetadataAuth, bondProvider, accessMethods, forceAccessUrl,
-        options = { usesAliasesForLocalizationPath: false }) {
+    constructor(providerName, sendMetadataAuth, bondProvider, accessMethods, forceAccessUrl, options) {
+        const defaultOptions = { usesAliasesForLocalizationPath: false };
+        this.options = { ...defaultOptions, ...options };
         this.providerName = providerName;
         this.sendMetadataAuth = sendMetadataAuth;
         this.bondProvider = bondProvider;
         this.accessMethods = accessMethods;
         this.forceAccessUrl = forceAccessUrl;
-        this.options = options;
     }
 
     accessMethodHavingSameTypeAs(accessMethod) {
--- a/martha/drs_providers.js
+++ b/martha/drs_providers.js
@@ -48,12 +48,14 @@ class AccessMethod {
 }
 
 class DrsProvider {
-    constructor(providerName, sendMetadataAuth, bondProvider, accessMethods, forceAccessUrl) {
+    constructor(providerName, sendMetadataAuth, bondProvider, accessMethods, forceAccessUrl,
+        options = { usesAliasesForLocalizationPath: false}) {
         this.providerName = providerName;
         this.sendMetadataAuth = sendMetadataAuth;
         this.bondProvider = bondProvider;
         this.accessMethods = accessMethods;
         this.forceAccessUrl = forceAccessUrl;
+        this.options = options;
     }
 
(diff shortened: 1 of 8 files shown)


martha/martha_v3.js
```diff
@@ -88,6 +88,12 @@ function getDrsFileName(drsResponse) {
     }
 }
 
+function getLocalizationPath(drsProvider, drsResponse) {
+    if (drsProvider.usesAliasesForLocalizationPath() && drsResponse && Array.isArray(drsResponse.aliases)) {
+        return drsResponse.aliases[0];
```
mcovarr, line 93: should there be any special checking here for the existence of an element 0 or `|| null` handling?

mcovarr (commented):
minor comment only, plus an integration test appears to have had its feathers ruffled

martha/drs_providers.js
```diff
@@ -48,12 +48,14 @@ class AccessMethod {
 }
 
 class DrsProvider {
-    constructor(providerName, sendMetadataAuth, bondProvider, accessMethods, forceAccessUrl) {
+    constructor(providerName, sendMetadataAuth, bondProvider, accessMethods, forceAccessUrl,
+        options = { usesAliasesForLocalizationPath: false }) {
```
jgainerdewar: I'm curious why you decided to use an options object here rather than a single boolean flag. This seems easy to trip over - if we ever wanted to supply a second option, that code would need to supply this one correctly as well.

martha/drs_providers.js
```diff
@@ -128,6 +130,15 @@ class DrsProvider {
         return this && overlapFields(requestedFields, MARTHA_V3_METADATA_FIELDS);
     }
 
+    /**
+     * This is hopefully a temporary measure until we can take the time to either get a new field
+     * added to the DRS spec or implement a temporary spec extension with the Terra Data Repo team.
+     * See BT-417 for more details.
+     */
+    usesAliasesForLocalizationPath() {
+        return this.options.usesAliasesForLocalizationPath;
```
jgainerdewar, line 139: Excuse my javascript naïveté (thanks autocorrect) - what happens here if `usesAliasesForLocalizationPath` isn't present in the `options` object? It seems easy to create a new provider for which that's the case.

martha/martha_v3.js
breilly2: I think this will be okay because `undefined` should become `null` when the Martha response is created, but I'll double check this.

martha/drs_providers.js
breilly2: I chose an `options` object here because this is truly an option with a reasonable default. You're right, though, that I didn't achieve my intended flexibility in the context of adding additional options. I'll fix that.

martha/martha_v3.js
breilly2: Yeah, `MarthaV3Response`'s constructor does `this.localizationPath = localizationPath || null` so an `undefined` `localizationPath` is OK. In fact, I had to modify test expectations to expect a `null` `localizationPath`.

martha/drs_providers.js
breilly2: I've made that harder now, but if it weren't in the `options` object, it would be `undefined`. Since `undefined` is falsy, any reasonable use should result in the same behavior as if it were explicitly `false`.

breilly2:
> an integration test appears to have had its feathers ruffled

This should be fixed in the current build from my PR feedback commit.

jgainerdewar (approved):
LGTM!

breilly2:
Jenkins retest, please. 🤞

mcovarr (approved):

PR #243, opened 2021-12-16
Add instructions for rotating the Martha prod deploy SA key [BT-460]

breilly2:
I've primed the process so that following these instructions should be fairly automatic and keep the process going.


jgainerdewar (approved):
LGTM! I really appreciate that you've created a process that will help us keep this current going forward.

aednichols (approved):

PR #244, opened 2022-01-18
Passports

dvoet:
Added support for `AccessUrlAuth.PASSPORT`. If the provider has this set, when an `accessUrl` is requested, martha will query externalcreds for a user's passport and make a `POST` request to get the `accessUrl`. If this fails, martha will use a fallback authentication mechanism which for Gen3 will be `AccessUrlAuth.FENCE_TOKEN`. In the process retries were added to `api_adapter.postJsonTo`.

For now a new drs provider was added called `PassportTestDrsProvider` which handles compact id `dg.test0`.

### DIFF
--- a/common/api_adapter.js
+++ b/common/api_adapter.js
@@ -31,7 +31,9 @@ async function getHeaders(url, authorization) {
 
 async function httpCallWithRetry(url, httpCall, retryAttempt = 1, delay = INITIAL_BACKOFF_DELAY) {
     try {
-        return await httpCall();
+        const response = await httpCall();
+        console.log(`Successfully received response from url '${url}'.`);
+        return response;
     } catch (error) {
         console.log(`Received error for url '${url}'. Attempt ${retryAttempt}.`);
         console.error(error);
@@ -59,20 +61,16 @@ async function httpCallWithRetry(url, httpCall, retryAttempt = 1, delay = INITIA
 
 function getJsonFrom(url, authorization) {
     return httpCallWithRetry(url, async () => {
-        const {body} = await get('get', url, authorization)
+        const {body} = await get('get', url, authorization);
         /*
          handle the case when Martha receives empty JSON body. The reason behind forming a response with status 500
          and throwing it in `try` is so that it can be caught locally and be retried.
          */
         if (Object.keys(body).length === 0) {
-            console.log(`Received an empty JSON body while trying to resolve url '${url}'. Attempt ${retryAttempt}. ` +
-                'Creating a response with status 500.');
-
-            const errorMsg = `Something went wrong while trying to resolve url '${url}'. It came back with empty JSON body!`;
+            const errorMsg = `Received an empty JSON body while trying to resolve url '${url}'`;
             throw new FailureResponse(500, errorMsg);
         }
         else {
-            console.log(`Successfully received response from url '${url}'.`);
             return body;
         }
     });
--- a/common/api_adapter.js
+++ b/common/api_adapter.js
@@ -29,9 +29,25 @@ async function getHeaders(url, authorization) {
     }
 }
 
-async function getJsonFrom(url, authorization, retryAttempt = 1, delay = INITIAL_BACKOFF_DELAY) {
+async function getJsonFrom(url, authorization) {
+    return httpCallWithRetry(url, async () => get('get', url, authorization))
+}
+
+async function postJsonTo(url, authorization, payload) {
+    return httpCallWithRetry(url, async () => {
+        const postReq = request.post(url, payload);
+        postReq.set('Content-Type', 'application/json');
+        if (authorization) {
+            postReq.set('authorization', authorization);
+        }
+
+        return postReq.then((response) => response.body);
+    })
+}
+
+async function httpCallWithRetry(url, httpCall, retryAttempt = 1, delay = INITIAL_BACKOFF_DELAY) {
     try {
-        const {body} = await get('get', url, authorization);
+        const {body} = await httpCall();
 
         /*
          handle the case when Martha receives empty JSON body. The reason behind forming a response with status 500
@@ -61,7 +77,7 @@ async function getJsonFrom(url, authorization, retryAttempt = 1, delay = INITIAL
 
                 return new Promise((resolve, reject) => {
                     setTimeout(() => {
-                        getJsonFrom(url, authorization, retryAttempt + 1, backOffDelay)
+                        httpCallWithRetry(url, httpCall, retryAttempt + 1, backOffDelay)
                             .then(resolve)
                             .catch((error) => { reject(error); });
                     }, delay);
@@ -71,16 +87,6 @@ async function getJsonFrom(url, authorization, retryAttempt = 1, delay = INITIAL
         }
         else { throw error; }
     }
-}
-
-function postJsonTo(url, authorization, payload) {
-    const postReq = request.post(url, payload);
-    postReq.set('Content-Type', 'application/json');
-    if (authorization) {
-        postReq.set('authorization', authorization);
-    }
-
-    return postReq.then((response) => response.body);
 }
 
 exports.get = get;
--- a/common/api_adapter.js
+++ b/common/api_adapter.js
@@ -29,22 +29,6 @@ async function getHeaders(url, authorization) {
     }
 }
 
-async function getJsonFrom(url, authorization) {
-    return httpCallWithRetry(url, async () => get('get', url, authorization))
-}
-
-async function postJsonTo(url, authorization, payload) {
-    return httpCallWithRetry(url, async () => {
-        const postReq = request.post(url, payload);
-        postReq.set('Content-Type', 'application/json');
-        if (authorization) {
-            postReq.set('authorization', authorization);
-        }
-
-        return postReq.then((response) => response.body);
-    })
-}
-
 async function httpCallWithRetry(url, httpCall, retryAttempt = 1, delay = INITIAL_BACKOFF_DELAY) {
     try {
         const {body} = await httpCall();
@@ -87,6 +71,22 @@ async function httpCallWithRetry(url, httpCall, retryAttempt = 1, delay = INITIA
         }
         else { throw error; }
     }
+}
+
+function getJsonFrom(url, authorization) {
+    return httpCallWithRetry(url, () => get('get', url, authorization));
+}
+
+function postJsonTo(url, authorization, payload) {
+    return httpCallWithRetry(url, () => {
+        const postReq = request.post(url, payload);
+        postReq.set('Content-Type', 'application/json');
+        if (authorization) {
+            postReq.set('authorization', authorization);
+        }
+
+        return postReq.then((response) => response.body);
+    });
 }
 
 exports.get = get;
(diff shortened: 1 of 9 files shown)


dvoet:
jenkins retest

martha/drs_providers.js
```diff
@@ -80,7 +82,7 @@ class DrsProvider {
             overlapFields(requestedFields, MARTHA_V3_ACCESS_ID_FIELDS) &&
             (this.forceAccessUrl || (accessMethod &&
                 this.accessMethods.find((m) => m.accessMethodType === accessMethod.type &&
-                    m.accessUrlAuth === AccessUrlAuth.FENCE_TOKEN &&
+                    (m.accessUrlAuth === AccessUrlAuth.FENCE_TOKEN || m.fallbackAccessUrlAuth === AccessUrlAuth.FENCE_TOKEN) &&
```
zarsky-broad: consider:
```suggestion
                    [m.accessUrlAuth, m.fallbackAccessUrlAuth].includes(AccessUrlAuth.FENCE_TOKEN) &&
```

common/api_adapter.js
```diff
@@ -29,25 +29,11 @@ async function getHeaders(url, authorization) {
     }
 }
 
-async function getJsonFrom(url, authorization, retryAttempt = 1, delay = INITIAL_BACKOFF_DELAY) {
+async function httpCallWithRetry(url, httpCall, retryAttempt = 1, delay = INITIAL_BACKOFF_DELAY) {
```
zarsky-broad, line 32: This only uses `url` for logging now, do we still want to pass it?

martha/drs_providers.js
```diff
@@ -114,6 +116,13 @@ class DrsProvider {
             overlapFields(requestedFields, MARTHA_V3_BOND_SA_FIELDS);
     }
 
+    shouldFetchPassports(accessMethod, requestedFields) {
+        return overlapFields(requestedFields, MARTHA_V3_ACCESS_ID_FIELDS) &&
+            accessMethod &&
+            this.accessMethods.find((m) => m.accessMethodType === accessMethod.type &&
+                m.accessUrlAuth === AccessUrlAuth.PASSPORT);
```
zarsky-broad, line 125: It's nuts to me that we don't have lodash or any other util library in this codebase

martha/martha_v3.js
```diff
@@ -347,12 +350,45 @@ async function retrieveFromServers(params) {
...
     // try doing just before we do it so that we can provide that detail in the error report.
     let hypotheticalErrorMessage;
 
+    const getAccessUrl = async (args) => {
+        const { providerAccessMethod: {accessUrlAuth, fallbackAccessUrlAuth}, httpsAccessUrl, accessToken, auth } = args;
+
+        if (accessUrlAuth === AccessUrlAuth.PASSPORT) {
+            let accessUrl;
+            try {
+                if (passports) {
```
zarsky-broad, line 363: I think this would be easier to read with the `if` wrapping the `try` instead of the current way

martha/martha_v3.js
```diff
@@ -347,12 +350,45 @@ async function retrieveFromServers(params) {
...
+                    accessUrl = await apiAdapter.postJsonTo(httpsAccessUrl, null, {"passports": passports});
```
zarsky-broad:
```suggestion
                    accessUrl = await apiAdapter.postJsonTo(httpsAccessUrl, null, {passports});
```

martha/martha_v3.js
```diff
@@ -347,12 +350,45 @@ async function retrieveFromServers(params) {
...
+                }
+            } catch (error) {
+                console.log(`Passport authorized request failed for ${httpsAccessUrl} with error ${error}, falling back to ${fallbackAccessUrlAuth} authorization`);
+            }
+            if (!accessUrl && fallbackAccessUrlAuth) {
```
zarsky-broad: Should this just go in the catch, and the predicate just becomes `if (fallbackAccessUrlAuth)`?

test/martha/martha_v3.test.js
```diff
@@ -153,6 +165,52 @@ test.serial('martha_v3 uses the default error handler for unexpected errors', as
     sinon.assert.callCount(response.send, 0);
 });
 
+test.serial('martha_v3 calls the correct endpoints when only the accessUrl is requested with passports', async (t) => {
+    const {
+        id: objectId, self_uri: drsUri,
+        access_methods: { 0: { access_id: accessId, access_url: { url: gcsUrl } } }
```
zarsky-broad: I think you want this:
```suggestion
        access_methods: [{ access_id: accessId, access_url: { url: gcsUrl } }]
```

test/martha/martha_v3.test.js
```diff
@@ -153,6 +165,52 @@ test.serial('martha_v3 uses the default error handler for unexpected errors', as
...
+    t.deepEqual(response.body, { accessUrl: drsAccessUrlResponse });
+
+    sinon.assert.callCount(getJsonFromApiStub, 3);
+    sinon.assert.callCount(postJsonToApiStub, 1);
+});
+
+test.serial('martha_v3 calls the correct endpoints when only the accessUrl is requested with passports but using fallback', async (t) => {
+    const {
+        id: objectId, self_uri: drsUri,
+        access_methods: { 0: { access_id: accessId, access_url: { url: gcsUrl } } }
```
zarsky-broad:
```suggestion
        access_methods: [{ access_id: accessId, access_url: { url: gcsUrl } }]
```

zarsky-broad (commented):
some thoughts on clariy

package-lock.json
```diff
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "martha",
```
breilly2, line 8: Not sure why this line would have changed... or anything in this file for that matter.

breilly2:
What is the new file `martha.http`? Was it added by accident?

test/martha/martha_v3.test.js
breilly2: This is just keeping consistent with other code like this in this file, which I added in the first place. I can't believe I didn't try that syntax, which makes me want to double-check that it works... though surely you know better than I do so this is probably just a 🤦 moment.

martha/martha_v3.js
breilly2: Can't do both this and your suggestion on line 372. My vote is to make this change instead of the other one, or leave it as-is.

martha/martha_v3.js
```diff
@@ -347,12 +350,45 @@ async function retrieveFromServers(params) {
...
     let localizationPath;
     let accessUrl;
+    let passports;
 
     // `fetch` below might time out in a way that we want to report as an error. Since the timeout
     // interrupts us while we're waiting for a response, we need to capture what we were about to
     // try doing just before we do it so that we can provide that detail in the error report.
     let hypotheticalErrorMessage;
 
+    const getAccessUrl = async (args) => {
```
breilly2: I think this function is doing a little too much by auto-handling the fallback case. Consequently:

* the caller has wasted time fetching a Fence token even if it didn't need to
* errors from calling Bond and Fence will cause the Martha invocation to fail even if the passport would have worked

I'd rather this function return nothing if fetching with the passport fails and move the fallback logic to the caller.

(I first noticed this from the tests; I was surprised to see that they expected Bond to be called.)

dvoet:
> What is the new file `martha.http`? Was it added by accident?

a mistake probably, I will remove it

martha/martha_v3.js
dvoet: I don't think this should be the caller's responsibility. This is one of the purpose's of Martha: figure out the authentication.

I agree that the caller's time is wasted fetching the fence token, I will work on that.

martha/martha_v3.js
dvoet: no, fallback needs to happen when fallbackAccessUrlAuth exists and there isn't an accessUrl yet. There are 2 reasons for not having an accessUrl 1) no passports, 2) an error getting the accessUrl using passports.

martha/martha_v3.js
dvoet: Unless you mean caller as in caller of `getAccessUrl` and not caller of Martha.

package-lock.json
zarsky-broad: this changes when running `npm install` with node 12's default npm

common/api_adapter.js
dvoet: the logs are important, let's keep it

martha/martha_v3.js
```diff
@@ -407,30 +469,31 @@ async function retrieveFromServers(params) {
...
-                try {
-                    const httpsAccessUrl = generateAccessUrl(drsProvider, urlParts, accessMethod.access_id);
-                    // Use the access token fetched in the call above or the auth submitted to Martha directly by the
-                    // caller as appropriate.
-                    const accessUrlAuth = drsProvider.determineAccessUrlAuth(accessMethod, accessToken, auth);
-                    console.log(`Requesting DRS access URL for '${url}' from '${httpsAccessUrl}'`);
-                    accessUrl = await apiAdapter.getJsonFrom(httpsAccessUrl, accessUrlAuth);
-                } catch (error) {
-                    throw new RemoteServerError(error, 'Received error contacting DRS provider.');
+                const fetchAccessUrl = async () => {
```
dvoet, line 476: I made this a function because repeated assignments of `accessUrl`, especially one inside an if block whose condition referenced `accessUrl` made the linter complain about race conditions.

dvoet:
@breilly2 @zarsky-broad I have replied to comments, PTAL

martha/martha_v3.js
```diff
@@ -347,12 +350,54 @@ async function retrieveFromServers(params) {
...
 
+    const getAccessUrl = async (accessUrlAuth, httpsAccessUrl, accessToken, auth) => {
+        if (accessUrlAuth === AccessUrlAuth.PASSPORT) {
+            if (passports) {
+                try {
+                    return await apiAdapter.postJsonTo(httpsAccessUrl, null, {passports});
+                } catch (error) {
+                    console.log(`Passport authorized request failed for ${httpsAccessUrl} with error ${error}`);
+                }
+            }
```
zarsky-broad:
```suggestion
        if (accessUrlAuth === AccessUrlAuth.PASSPORT && passports) {
            try {
                return await apiAdapter.postJsonTo(httpsAccessUrl, null, {passports});
            } catch (error) {
                console.log(`Passport authorized request failed for ${httpsAccessUrl} with error ${error}`);
            }
```

martha/martha_v3.js
breilly2: Yes, sorry, I meant the caller in the context of this function, `getAccessUrl`, not the caller of Martha.

zarsky-broad (approved):
lgtm from a code perspective, one potential simplification.

breilly2 (approved):
👍  Thanks for addressing feedback.

The changes in the last round aren't exactly what I expected. Specifically, I didn't expect `shouldFetchFenceAccessToken` to need to change. However, the tests look correct now, and I'm sure there's good reason for the way things are that aren't immediately apparent to me.

martha/martha_v3.js
dvoet: This is actually not the same because when there is no passport the final `else` statement at the end of the function will throw an error. But I am going to change this back to a switch.

dvoet:
jenkins retest

dvoet:
jenkins retest

dvoet:
jenkins retest

dvoet:
jenkins retest

PR #245, opened 2022-01-20
Update request logging [BT-508]

breilly2:


jgainerdewar:
Confirmed that this log line no longer includes anything it shouldn't 🎉 

However, when I tested with a bogus token, the errors printed in my logs did contain that token - when we get a RemoteServerError it contains the full request, and we print the full error object in the logs. I think we need to update `helpers.logAndSendServerError` as well.

breilly2:
@jgainerdewar @mcovarr Back to ready for review. (or not... looks like some tests are failing now 😞 )

mcovarr (approved):
lgtm

jgainerdewar (approved):
Retested and confirmed no tokens in logs. 🎉

PR #247, opened 2022-01-27
Enable signed URL's for CRDC [BT-269] [BT-541]

jgainerdewar:
I would look at the first two commits one at a time, I did a bunch of renaming that will clutter up the summary diff. 

These tests pass, but I'd appreciate a sanity check on whether I told them to expect the right behavior!


jgainerdewar:
jenkins retest

breilly2 (approved):
👍  Looks good to me!

mcovarr (approved):
LGTM delta the minor conflict 👍

PR #248, opened 2022-02-01
Enable BDC signed URLs [BT-267]

breilly2:
Note: I removed a test that was redundant with the one I had to update. These used to be slightly different but they converged several commits ago. Fortunately, the behavior is covered by the recent merging of Khalid's long-open PR, so the test I removed is truly redundant.

### DIFF
--- a/martha/drs_providers.js
+++ b/martha/drs_providers.js
@@ -160,7 +160,6 @@ class BioDataCatalystDrsProvider extends DrsProvider {
             MetadataAuth.NO,
             BondProvider.FENCE,
             [
-                //  BT-236 BDC access (signed) URLs temporarily turned off
                 new AccessMethod(AccessMethodType.GCS, AccessUrlAuth.FENCE_TOKEN, FetchAccessUrl.YES)
             ],
             forceAccessUrl);
--- a/martha/drs_providers.js
+++ b/martha/drs_providers.js
@@ -161,7 +161,7 @@ class BioDataCatalystDrsProvider extends DrsProvider {
             BondProvider.FENCE,
             [
                 //  BT-236 BDC access (signed) URLs temporarily turned off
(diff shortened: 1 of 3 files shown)


martha/drs_providers.js
```diff
@@ -159,7 +159,7 @@ class BioDataCatalystDrsProvider extends DrsProvider {
             BondProvider.FENCE,
             [
                 //  BT-236 BDC access (signed) URLs temporarily turned off
```
jgainerdewar: Remove this comment?

jgainerdewar (commented):
Code looks good to me, I see some CI errors so will wait until those are resolved to 👍

breilly2:
Yeah, there are currently problems with fiab-start/stop, so I'm not surprised that the Jenkins build failed.

martha/drs_providers.js
breilly2: Oh, I'm glad you mentioned this! Searching the code for BT-236 also shows that there are some skipped tests flagged with this ticket that I'll need to address as well.

jgainerdewar (approved):
LGTM!

test/martha/martha_v3.test.js
```diff
@@ -1124,6 +1106,8 @@ test.skip('martha_v3 should return 500 if access method parsing fails', async (t
 
 /* BT-236 Skip testing access token fetch failure since that is not something this code even attempts with BDC
  * signed URLs turned off. */
```
mcovarr, line 1151: is this comment (and the copy below) stale now that BDC signed URLs are turned on?

breilly2: Sorta. I was leaving it to provide context for the comment that I added. I guess I could have re-written it instead, and I know it could be read from git history. I can make changes if you want.

breilly2:
Jenkins retest please

mcovarr (approved):

PR #249, opened 2022-02-09
POL-77 mTLS

dvoet:
add support for mTLS with secrets pulled from Google Secret Manager

see also https://github.com/broadinstitute/terraform-ap-deployments/pull/535

### DIFF
--- a/martha/martha_v3.js
+++ b/martha/martha_v3.js
@@ -640,3 +640,4 @@ exports.generateAccessUrl = generateAccessUrl;
 exports.getHttpsUrlParts = getHttpsUrlParts;
 exports.overridePencilsDownSeconds = overridePencilsDownSeconds;
 exports.PROTOCOL_PREFIX_DRS = PROTOCOL_PREFIX_DRS;
+exports.secretManagerServiceClient = secretManagerServiceClient;
--- a/martha/martha_v3.js
+++ b/martha/martha_v3.js
@@ -376,7 +376,7 @@ async function retrieveFromServers(params) {
             case AccessUrlAuth.PASSPORT:
                 if (passports) {
                     try {
-                        let clientPrivateKey, clientCert;
+                        let clientCert, clientPrivateKey;
                         if (drsProvider.clientCertSecretName && drsProvider.clientPrivateKeySecretName) {
                             clientPrivateKey = await getSecret(drsProvider.clientPrivateKeySecretName);
                             clientCert = await getSecret(drsProvider.clientCertSecretName);
--- a/martha/martha_v3.js
+++ b/martha/martha_v3.js
@@ -19,6 +19,8 @@ const {
     AccessUrlAuth
 } = require("./drs_providers");
 
+const {SecretManagerServiceClient} = require('@google-cloud/secret-manager');
+
 const config = require('../common/config');
 const apiAdapter = require('../common/api_adapter');
 const url = require('url');
@@ -43,6 +45,17 @@ let pencilsDownSeconds = 58;
 const overridePencilsDownSeconds = (seconds) => {
     pencilsDownSeconds = seconds;
 };
+
+const secretManagerServiceClient = new SecretManagerServiceClient();
+
+async function getSecret(name) {
+    const [version] = await secretManagerServiceClient.accessSecretVersion({
+        name: name,
+    });
+
+    // Extract the payload as a string.
+    return version.payload.data.toString();
+}
 
 /**
  * Returns the first access method in `drsProvider.accessMethodTypes()` with a type that matches the type of an access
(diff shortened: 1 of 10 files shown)


snwessel (approved):
Makes sense to me

TomConner (approved):

TomConner (approved):

dinvlad (approved):

test/common/config_test.js
```diff
@@ -91,7 +99,7 @@ test('config parseConfigJson should return an empty json in ENV_MOCK', (t) => {
     t.deepEqual(configJson, {});
 });
 
-test('config parseConfigJson should return an empty json for a missing temp file in ENV_DEV', (t) => {
+test('config parseConfigJson should return an empty json for aanswer for the Cromwell dev environment missing temp file in ENV_DEV', (t) => {
```
breilly2: I don't understand what this is trying to say. Maybe this is an unintentional change? At a minimum, there's an extra "a".

test/martha/martha_v3.test.js
```diff
@@ -171,11 +176,16 @@ test.serial('martha_v3 calls the correct endpoints when only the accessUrl is re
...
+    const clientCert = "client cert";
+
     const drs = drsUrls(config.HOST_PASSPORT_TEST, objectId, accessId);
     const drsAccessUrlResponse = mockGcsAccessUrl(gcsUrl);
     getJsonFromApiStub.withArgs(drs.objectsUrl, null).resolves(passportTestResponse);
     getJsonFromApiStub.withArgs(ecmUrls('ras').passportUrl, terraAuth).resolves(passport);
-    postJsonToApiStub.withArgs(drs.accessUrl, null, {"passports": [passport]}).resolves(drsAccessUrlResponse);
+    postJsonToApiStub.withArgs(drs.accessUrl, null, {"passports": [passport]}, clientKey, clientCert).resolves(drsAccessUrlResponse);
+    accessSecretVersionStub.withArgs({name: config.rasClientMTLSKeySecretName}).resolves([{payload: {data: clientKey}}]);
+    accessSecretVersionStub.withArgs({name: config.rasClientMTLSCertSecretName}).resolves([{payload: {data: clientCert}}]);
```
breilly2, line 188: Should add this to the end of this test:
```
    sinon.assert.callCount(accessSecretVersionStub, 2);
```
Otherwise the test will pass even if these calls don't happen.

I had experimented with using the sinon API to declare these stub calls to be required, but it wasn't working for some reason that I couldn't figure out. The best compromise I came up with was to assert the number of expected calls at the end of the test function. It's not perfect because the test _could_ still pass if there are 2 calls but they don't match these calls. However, if the test is built so that the results of these stubs is necessary for the rest of the assertions, then call count should be good enough.

martha/martha_v3.js
```diff
@@ -44,6 +46,17 @@ const overridePencilsDownSeconds = (seconds) => {
     pencilsDownSeconds = seconds;
 };
 
+const secretManagerServiceClient = new SecretManagerServiceClient();
+
+async function getSecret(name) {
+    const [version] = await secretManagerServiceClient.accessSecretVersion({
```
breilly2, line 52: Is this pretty quick? (We're calling this twice on every execution that uses passports.) Is there any API rate limit that we need to be worried about? I ask because we could potentially fetch these on cloud function initialization (or lazy fetch and cache) in case the instance gets used more than once.

breilly2 (commented):
A couple minor points/questions.

martha/martha_v3.js
dvoet: 90k per minute is the quota which is well beyond our spikes.
https://cloud.google.com/secret-manager/quotas

test/common/config_test.js
dvoet: probably an accidental paste

test/martha/martha_v3.test.js
dvoet: I am inclined not to do this. Does the test care that these were called or does it only care that the post stubbed on the line above was called with the right key and cert? I think we care about the latter and the test as written will fail if the key and cert are not right.

Put another way, if we decide we need to cache the key and cert and these apis are not called the same way, do we want this test to fail? I don't think so.

breilly2: Ahh, okay. My worry was that the test could succeed if Martha didn't use the key/cert. However, now I see that the stub for `POST` to `drs.accessUrl` requires these, and if Martha doesn't have and use the key/cert for that call then that stub won't match and the response assertion will fail.

I'll downgrade this to a minor nit: putting these stub definitions before the stub for `POST` to `drs.accessUrl` makes the stub flow more closely match the execution order which I think clarifies how everything is connected.

breilly2 (approved):
👍  Downgraded my change request to a minor nit. Change would be appreciated, but no re-review required either way.

PR #256, opened 2022-04-14
Update Martha release process [BT-607]

jgainerdewar:
This is done as part of an effort to standardize the release process for non-monolith products.

### DIFF
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -6,6 +6,7 @@ Martha's release process is documented and tracked through a Jira checklist.
 Martha is updated rarely enough that we typically include a release in the 
 work of each ticket. Navigate to the ticket associated with the work you're 
 releasing, or create a new ticket for the release. In the Jira ticket, click 
-on Add Checklist, then click on the button with three dots in the upper right 
+on Add Checklist (may just be a square both with a check in it), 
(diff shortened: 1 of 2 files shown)


docs/release-process.md
```diff
@@ -0,0 +1,11 @@
+Martha Release Process
+=========
+
+Martha's release process is documented and tracked through a Jira checklist. 
+
+Martha is updated rarely enough that we typically include a release in the 
+work of each ticket. Navigate to the ticket associated with the work you're 
+releasing, or create a new ticket for the release. In the Jira ticket, click 
+on Add Checklist, then click on the button with three dots in the upper right
```
breilly2: "Add Checklist" is this button, right?
Screen Shot 2022-04-14 at 3 07 52 PM

breilly2 (approved):
👍  Assuming I correctly identified the "Add Checklist" button. It might be worth adding more description about how to find this button; it wasn't immediately obvious to me.

docs/release-process.md
jgainerdewar: Yes! In other views of the ticket there's text in it. I'll expand the description here.

mspector (approved):
Seems clear to me!

PR #257, opened 2022-06-07
BT-595 Updates to instructions for rotating deploy key

jgainerdewar:


mspector (approved):
looks good!

docs/deployment-key-rotation.md
```diff
@@ -13,14 +13,15 @@ The service account key used to deploy Martha needs to be rotated every 90 days.
 ```
 vault write secret/dsde/martha/prod/deploy-account.json @[key .json file]
 ```
-5. Disable or delete the existing key.
+5. Disable or delete the existing key (id found on above page). Using your `@firecloud.org` account:
 ```
+gcloud config set project broad-dsde-prod
```
breilly2, line 18: I can't say for sure, but I don't remember ever having to do this. _Maybe_ it's because I had this config already set, but I'm usually good at avoiding having any persistent config like this pointing at prod environments.

breilly2 (approved):

docs/deployment-key-rotation.md
jgainerdewar: Luckily still had this still hanging out in my Terminal buffer, so I can tell you that the first time I ran the below command (logged in with my FC account) I got: 
```
ERROR: (gcloud.iam.service-accounts.keys.disable) PERMISSION_DENIED: Permission iam.serviceAccountKeys.disable is required to perform this operation on service account key projects/broad-dsde-cromwell-dev/serviceAccounts/cloud-functions-account@broad-dsde-prod.iam.gserviceaccount.com/keys/<key id>.`
```

Note the incorrect project name in there. I set the project and ran it again, and it worked. It seems like project selections are account-specific, maybe you do have the prod project set for your FC account, since that's all it's ever used for?

PR #260, opened 2022-07-19
[BT-692] Fix doc link

breilly2:
Linking to BT-692 because that's what I was working on when I discovered this minor doc bug.


jgainerdewar (approved):
Nice catch!

kraefrei (approved):

PR #261, opened 2022-08-01
[BT-706] sample refactors

mspector:


test/martha/forceAccessUrl.test.js
```diff
@@ -0,0 +1,68 @@
...
+
+const { marthaV3Handler: marthaV3 } = require('../../martha/martha_v3');
+const apiAdapter = require('../../common/api_adapter');
+
+const terraAuth = "Bearer XYZ_TerraAuthToken";
+const bucketId = "bucket_id";
+const responseId = `v1_ResponseId_${bucketId}`;
+const gsAccessId = "gcp-us-central1";
+const gsUrl = `gs://broad-jade-dev-data-bucket/BucketIdPrefix/${bucketId}`;
+const drsUri = `drs://${config.HOST_TDR_DEV}/${responseId}`
```
mspector, line 14: `config.HOST_TDR_DEV` appears to be required for this test to pass. Is this really a unit test after all?

test/martha/forceAccessUrl.test.js
```diff
@@ -0,0 +1,68 @@
...
+
+
+const jadeAccessUrlMetadataResponse = {
```
mspector, line 17: in the original test suite, this object is [defined in an all-encompassing "resources" file](https://github.com/broadinstitute/martha/blob/dev/test/martha/_martha_v3_resources.js#L163).

In this test, it appears to need far fewer fields than provided in that resources file. The definition here demonstrates the essential fields.

Is this an example of DRY gone too far? We may have to repeat versions of this object in other tests, but keeping it small/constrained in each test makes it easier to read.

test/martha/_helpers.js
```diff
@@ -0,0 +1,58 @@
+const sinon = require('sinon');
+
+const {
+    MARTHA_V3_ALL_FIELDS,
+} = require("../../martha/martha_fields");
+
+
+const mockRequest = (req, options={}, auth="Bearer XYZ_TerraAuthToken") => {
```
mspector, line 8: the original implementation of this helper function [is placed inside of the main testing code](https://github.com/broadinstitute/martha/blob/dev/test/martha/martha_v3.test.js#L68), and relies on global variables defined there.

Helpers should be small, pure (in the functional sense), and be separated in another file/namespace. (Namespaces help with cognitive overhead)

test/martha/forceAccessUrl.test.js
```diff
@@ -0,0 +1,68 @@
+const test = require('ava');
+const sinon = require('sinon');
+const config = require('../../common/config');
+const helpers = require('./_helpers.js')
+
+const { marthaV3Handler: marthaV3 } = require('../../martha/martha_v3');
+const apiAdapter = require('../../common/api_adapter');
+
+const terraAuth = "Bearer XYZ_TerraAuthToken";
```
mspector, line 9: In the original tests, these variables were extracted from [the all-encompassing "resources" file](https://github.com/broadinstitute/martha/blob/dev/test/martha/_martha_v3_resources.js) by way of a destructuring assignment.

Defining variables upfront and then creating a new object (vs. extracting variables from a "god config") feels more straightforward. What arguments are there against this approach?

test/martha/forceAccessUrl.test.js
```diff
@@ -0,0 +1,68 @@
...
+  "access_methods": [
+    {
+      "type": "gs",
+      "access_id": `${gsAccessId}`,
+    },
+  ],
+};
+
+
+test.serial.beforeEach(() => {
```
mspector, line 28: I'm not enough of an expert with `ava` (the test runner) to know if defining this here will affect tests outside of this file. But it does seem like each semantically-related group of tests should be able to have their own set of setup/teardown functionality.

test/martha/forceAccessUrl.test.js
```diff
@@ -0,0 +1,68 @@
...
+
+test.serial('(custom) martha_v3 calls the correct endpoints when access url fetch is forced for TDR', async (t) => {  
+    const drs = helpers.drsUrls(
+      config.HOST_TDR_DEV, 
+      responseId,
+      gsAccessId,
+    );
+    
+    // 2021-10-04 Jade returns `"headers": null` if there are no headers, 
+    // while the Gen3 repos we have worked with to date omit the "headers" kv completely.
```
mspector, line 42: This comment was brought over from the original tests. It is a great example of a constructive comment.

test/martha/forceAccessUrl.test.js
```diff
@@ -0,0 +1,68 @@
...
+    const drsAccessUrlResponse = {...helpers.mockGcsAccessUrl(gsUrl), 'headers': null};
+
+    getJsonFromApiStub.withArgs(drs.objectsUrl, terraAuth).resolves(jadeAccessUrlMetadataResponse);
+    getJsonFromApiStub.withArgs(drs.accessUrl, terraAuth).resolves(drsAccessUrlResponse);
+
+    const request = helpers.mockRequest(
+      {
```
mspector, line 49: Formatting matters for readability. The test suite uses `eslint`, but it should be configured to format code in a consistent style as well as check for errors.

PR #263, opened 2022-11-17
WX-839: Add Strict-Transport-Security and CacheControl headers to Martha responses

JVThomas:
Addresses [WX-839](https://broadworkbench.atlassian.net/browse/WX-839)

PR adds the state two security headers to Martha responses. Responses were adjusted per route since functions-framework doesn't allow for custom middleware out of the box.

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,7 +12,7 @@ const getSignedUrlV1 = require('./handlers/getSignedUrlV1');
 const addSecurityHeaders = (res) => {
     res.set({
         'Strict-Transport-Security': 'max-age=63072000',
-        'CacheControl': 'no-store',
+        'Cache-Control': 'no-store',
     });
 };
 
--- a/index.js
+++ b/index.js
@@ -9,19 +9,30 @@ const { marthaV3Handler } = require('./martha/martha_v3');
 const { fileSummaryV1Handler } = require('./fileSummaryV1/fileSummaryV1');
 const getSignedUrlV1 = require('./handlers/getSignedUrlV1');
 
+const addSecurityHeaders = (res) => {
+    res.set({
+      'Strict-Transport-Security': 'max-age=63072000',
+      CacheControl: 'no-store',
+    });
+}
+
 exports.martha_v2 = (req, res) => {
+    addSecurityHeaders(res);
     corsMiddleware(req, res, () => marthaV2Handler(req, res));
 };
 
 exports.martha_v3 = (req, res) => {
+    addSecurityHeaders(res);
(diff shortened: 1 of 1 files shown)


jgainerdewar:
jenkins retest

aednichols:
@jgainerdewar that Jenkins job fails because the FiaB allocator (and FiaBs in general) have been removed. Justin is going to look into whether that test adds much value on top of the existing unit tests and if not, delete it.

jgainerdewar:
Ah! I was wondering if that was the problem. Sounds like a good plan... for Justin :)

aednichols:
Ugh I managed to avoid that mistake so far but I knew it would happen eventually!

jgainerdewar (approved):

jgainerdewar:
As discussed on Slack, with the coming retirement of Martha and the manual integration tests we run at release time, we don't think it's worth it to port these Jenkins tests to BEEs.

index.js
```diff
@@ -9,19 +9,30 @@ const { marthaV3Handler } = require('./martha/martha_v3');
 const { fileSummaryV1Handler } = require('./fileSummaryV1/fileSummaryV1');
 const getSignedUrlV1 = require('./handlers/getSignedUrlV1');
 
+const addSecurityHeaders = (res) => {
+    res.set({
+        'Strict-Transport-Security': 'max-age=63072000',
```
aednichols, line 14: I see that https://hstspreload.org/#deployment-recommendations advises starting with a short age, verifying that it works, and later ramping up to two years. Was this taken into consideration?

JVThomas: I tested with `Postman` to see if the header was included in the responses when running Martha locally (which it did). I can't actually test to see if the incoming requests are being sent via `https` since localhost only runs on the `http` protocol.

Aside: The deployment steps are a recommendation for being included in Google's preload list, which is a nice to have, but isn't considered official spec and is considered optional.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security#preloading_strict_transport_security

I didn't see the `preload` directive listed in the ticket details, so I assumed it could be ignored.

aednichols: FYI we have a convention as well as actual certificates in place (in some places) that allow `local.broadinstitute.org` for HTTPS testing. And, alias that to `localhost` in the `hosts`.

aednichols (approved):
Code looks good to me.

I created a companion PR https://github.com/broadinstitute/dsp-jenkins/pull/646 to disable the obsolete tests.

PR #264, opened 2022-12-01
[ID-351] enable signed urls for TDR

Shakespeared:
Signed URLs are already turned on for Azure workspaces, but now that TDR is capable of serving signed URLs for data in Google requester pays buckets, we can enable it for Google as well. This one line change just flips the switch.

### DIFF
--- a/martha/drs_providers.js
+++ b/martha/drs_providers.js
@@ -172,7 +172,7 @@ class TerraDataRepoDrsProvider extends DrsProvider {
             MetadataAuth.YES,
             BondProvider.NONE,
             [
-                new AccessMethod(AccessMethodType.GCS, AccessUrlAuth.CURRENT_REQUEST, FetchAccessUrl.NO),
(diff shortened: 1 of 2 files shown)


jgainerdewar:
It looks like there are some tests that need to be updated here. 

We're planning to release Martha early next week, hopefully on Monday, for a security patch. If this merges soon it can hitch a ride!

Shakespeared:
> It looks like there are some tests that need to be updated here.
> 
> We're planning to release Martha early next week, hopefully on Monday, for a security patch. If this merges soon it can hitch a ride!

Yep, I realized right after I made this that there's test updates. Oh that's great, I'll make those test updates right away, if this could get in the release on Monday that would be amazing.

andy7i (approved):
lgtm 👍

Ghost-in-a-Jar (approved):

jgainerdewar (approved):
LGTM!

PR #276, opened 2023-03-31
[ID-487] update TDR host URL values in config

Shakespeared:
It was discovered in testing that some of the host URLs in the configs weren't defaulting to the correct values in certain environments. This is an update to fix the config values to be in line with expected behavior. 

Ticket https://broadworkbench.atlassian.net/browse/ID-487

slack thread for context: https://broadinstitute.slack.com/archives/C01KXE68J5D/p1680181716980209?thread_ts=1679508094.938999&cid=C01KXE68J5D


tlangs (approved):

common/config.js
```diff
@@ -17,11 +19,11 @@ const HOST_THE_ANVIL_PROD = 'gen3.theanvil.io';
 const HOST_THE_ANVIL_STAGING = 'staging.theanvil.io';
 const HOST_TERRA_DATA_REPO_PROD = 'data.terra.bio';
 const HOST_TERRA_DATA_REPO_STAGING = 'staging.data.terra.bio';
```
mbaumann-broad: TDR is also deployed in the Alpha tier here: https://data.alpha.envs-terra.bio
Therefore the following constant should be added here:
```
const HOST_TERRA_DATA_REPO_ALPHA = 'data.alpha.envs-terra.bio';

And, I believe, the new constant should be added to `configExport` at the end of this file.

File: common/config.js [hunk 2]
```diff
@@ -131,6 +133,12 @@ function configDefaultsForEnv({marthaEnv, dsdeEnv = dsdeEnvFrom(marthaEnv)}) {
                         return HOST_MOCK_DRS;
                     case ENV_PROD:
                         return HOST_TERRA_DATA_REPO_PROD;
+                    case ENV_STAGING:
+                    case ENV_ALPHA:
```
mbaumann-broad: The TDR deployment must be in the same tier as Sam (and Martha, etc.) to work. This is because passing the auth info (i.e., token) from Sam in one tier to TDR (or any other Terra service) in another tier will fail.

The `terraDataRepoHost` block must map each supported tier to the TDR instance in that same tier, as shown below:
```
        terraDataRepoHost:
            (() => {
                switch (marthaEnv) {
                    case ENV_MOCK:
                        return HOST_MOCK_DRS;
                    case ENV_PROD:
                        return HOST_TERRA_DATA_REPO_PROD;
                    case ENV_STAGING:
                        return HOST_TERRA_DATA_REPO_STAGING;
                    case ENV_ALPHA:
                        return HOST_TERRA_DATA_REPO_ALPHA;
                    case ENV_DEV:
                    case ENV_CROMWELL_DEV:
                        return HOST_TDR_DEV;
                    default:
                        console.log(`Unsupported environment: '${marthaEnv}'.`);
                }
            })(),
  ```
I meant to put the above in the Slack message [here](https://broadinstitute.slack.com/archives/C01KXE68J5D/p1680181716980209?thread_ts=1679508094.938999&cid=C01KXE68J5D), yet I see I missed an update there and I understand the confusion. Hopefully this is clear now. 

Because all known/supported tiers are explicitly mapped to TDR in that same tier, I think the only reasonable `default` action is to issue a message indicating something is amiss.
I see `console.log` used for this elsewhere in the Martha code, so hopefully, that is appropriate here, too(?)

**Review by mbaumann-broad — CHANGES_REQUESTED** [review]
Requesting a small change so TDR for the Alpha tier is handled the same as the other tiers are.

File: common/config.js (hunk 2 above)
Comment (Shakespeared) [reply common/config.js]: Ah yeah I see I didn't quite get the nuance of that from your earlier comment, but this makes sense! I'll wait for you to finish reviewing then I'll update with changes. Thank you!

File: common/config.js [hunk 3]
```diff
@@ -17,11 +19,12 @@ const HOST_THE_ANVIL_PROD = 'gen3.theanvil.io';
 const HOST_THE_ANVIL_STAGING = 'staging.theanvil.io';
 const HOST_TERRA_DATA_REPO_PROD = 'data.terra.bio';
 const HOST_TERRA_DATA_REPO_STAGING = 'staging.data.terra.bio';
```
mbaumann-broad: The TDR staging hostname `staging.data.terra.bio` is incorrect. The correct TDR staging hostname is `data.staging.envs-terra.bio`

Please change from this:
```
const HOST_TERRA_DATA_REPO_STAGING = 'staging.data.terra.bio';
```
to this:
```
const HOST_TERRA_DATA_REPO_STAGING = 'data.staging.envs-terra.bio';
```

mbaumann-broad (changes requested):
Overall, looks good based on testing!

One small yet important change requested.

mbaumann-broad (approved):
Looks great - thank you!

bstitson-broad (approved):

common/config.js
```diff
@@ -131,8 +134,15 @@ function configDefaultsForEnv({marthaEnv, dsdeEnv = dsdeEnvFrom(marthaEnv)}) {
                         return HOST_MOCK_DRS;
                     case ENV_PROD:
                         return HOST_TERRA_DATA_REPO_PROD;
-                    default:
+                    case ENV_STAGING:
                         return HOST_TERRA_DATA_REPO_STAGING;
+                    case ENV_ALPHA:
+                        return HOST_TERRA_DATA_REPO_ALPHA;
+                    case ENV_DEV:
```
Ghost-in-a-Jar, line 141: It is ok that this one is blank?

Ghost-in-a-Jar (approved):
Minor comment

PR #282, opened 2023-07-20
ID-694 Fix Integration Tests

tlangs:
Because Bond now requires a `state` parameter to avoid CSRF attacks, the Bond API endpoints called by the integration tests changed. Additionally, the `state` now means that integration test suites need to be run separately, since getting state for one integration suite will invalidate the state for the other.


tlangs:
Merging to fix integration tests. I've run them locally, and they all pass now! No app changes needed for approval.

